PR #68, opened 2018-08-06
Move case reference generation into own method

zuzak:


nickmcmahonHOD:
this.timestamp = LocalDateTime.now();
this.reference = String.format("%s/%07d/%ty", this.type, caseNumber, this.timestamp);

%ty m9

PR #75, opened 2018-08-29
Save Corerspondent and Reference

nickmcmahonHOD:


src/main/java/uk/gov/digital/ho/hocs/casework/casedetails/model/ReferenceType.java
```diff
@@ -4,10 +4,13 @@
 
 public enum ReferenceType {
 
-   MP("MP"),
+    MEMBER_REFERENCE("Member Reference"),
    HO_REFERENCE("HO Reference"),
    PASSPORT_NUMBER("Passport Number"),
```
zuzak: Your indentation's wonky

PR #135, opened 2018-11-29
added properties and updated configuration

nwoodHO:


src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
```diff
@@ -18,17 +19,20 @@
...
+    public CaseDataService(CaseDataRepository caseDataRepository, ObjectMapper objectMapper, AuditClient auditClient) {
         this.caseDataRepository = caseDataRepository;
+        this.auditClient = auditClient;
         this.objectMapper = objectMapper;
     }
 
     @Transactional
     public CaseData createCase(CaseDataType caseDataType, Map<String, String> data) {
         CaseData caseData = new CaseData(caseDataType, caseDataRepository.getNextSeriesId(), data, objectMapper);
+        auditClient.createCaseAudit(caseData);
```
SWhitlam: We should audit after something has happened rather than before.

src/test/java/uk/gov/digital/ho/hocs/casework/api/CaseDataServiceTest.java
```diff
@@ -59,6 +62,17 @@ public void shouldCreateCaseWithValidParamsNullData() throws EntityCreationExcep
         verifyNoMoreInteractions(caseDataRepository);
     }
 
+    @Test
+    public void shouldAuditCreateCase() throws EntityCreationException {
```
SWhitlam: Could do with a test around what happens if the audit client fails

PR #205, opened 2019-03-06
Case By Reference

nickmcmahon01:


src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataResource.java
```diff
@@ -32,6 +35,13 @@ public CaseDataResource(CaseDataService caseDataService) {
         return ResponseEntity.ok(CreateCaseResponse.from(caseData));
     }
 
+    @GetMapping(value = "/case/r/{reference:[A-Z]{2,}%2F[0-9]{7}%2F[0-9]{2}}")
+    ResponseEntity<GetCaseResponse> getCase(@PathVariable String reference) throws UnsupportedEncodingException {
+        String decodedRef = URLDecoder.decode(reference, StandardCharsets.UTF_8.name());
```
nickmcmahon01: I'm not sure if we should be throwing this here, it should never be thrown because StandardCharsets.UTF_8.name() is hardcoded not passed in, so if it works once it should 'just work' all the time

PR #241, opened 2019-04-09
use @Async and @Retry on void audit and notify calls.

nickmcmahon01:
Don't cache nominated contacts because we can afford the lookup due to the Async


src/main/java/uk/gov/digital/ho/hocs/casework/application/CacheWarmer.java
```diff
@@ -69,7 +67,7 @@ private void primeStandardLinesCache(){
                 this.infoClient.populateStandardLine(getStandardLineResponse.getTopicUUID(), getStandardLineResponse);
             }
         } catch(Exception e) {
-            log.warn("Failed to prime StandardLines. :" + e.toString(), value(EVENT, CACHE_PRIME_FAILED));
+            log.warn("Failed to prime StandardLines. : {}", e.toString(), value(EVENT, CACHE_PRIME_FAILED));
```
SWhitlam: These should look as the exception value rather than as an inline string. 
e.g. log.warn("Failed to prime StandardLines.", value(EVENT, CACHE_PRIME_FAILED), value(EXCEPTION, value(e.toString());

src/main/java/uk/gov/digital/ho/hocs/casework/application/ContextAwareExecutorDecorator.java
```diff
@@ -0,0 +1,45 @@
...
+        this.executor = executor;
+    }
+
+    @Override
+    public void execute(Runnable command) {
+        Runnable ctxAwareCommand = decorateContextAware(command);
+        executor.execute(ctxAwareCommand);
+    }
+
+    private Runnable decorateContextAware(Runnable command) {
```
SWhitlam: Not sure what this is doing, it looks like it will clear the main thread MDC then replace it again.

src/main/java/uk/gov/digital/ho/hocs/casework/application/CacheWarmer.java
nickmcmahon01: I agree, ill do this a separate piece for the whole service

src/main/java/uk/gov/digital/ho/hocs/casework/application/ContextAwareExecutorDecorator.java
nickmcmahon01: I (blindly) followed this guide: http://www.chrisport.ch/java/2016/10/12/java-spring-context-aware.html

PR #260, opened 2019-08-13
HOCS-1007 - Broken Access Controls

hubert-raich:
-routing all UI document requests via casework service to make use of a case permission model


src/test/java/uk/gov/digital/ho/hocs/casework/api/CaseDocumentServiceTest.java
```diff
@@ -0,0 +1,93 @@
...
+    private CaseDocumentService caseDocumentService;
+
+    @Before
+    public void setUp() {
+        caseDocumentService = new CaseDocumentService(documentClient);
+    }
+
+    @Test
+    public void getDocuments() {
+        GetDocumentsResponse documentsResponse = mock(GetDocumentsResponse.class);
```
Paulcr-ho: It is best to instantiate actual instances of GetDocumentResponse, DocumentDto and S3Document in these tests, so that the values can be verified after the call to the service.

src/test/java/uk/gov/digital/ho/hocs/casework/client/documentclient/DocumentClientTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+    @Before
+    public void setUp() {
+        documentClient = new DocumentClient(restHelper, documentService);
+    }
+
+    @Test
+    public void getDocuments() {
+        GetDocumentsResponse mockedResult = mock(GetDocumentsResponse.class);
+        String url = "/document/reference/" + caseUUID.toString();
```
Paulcr-ho: It's best to create actual instances of GetDocumentResponse, DocumentDto and S3Document in these tests. The actual instance values can then be checked after the client call.
An example of this if the Http Headers instantiation in getFile of RestHelperTest.

PR #291, opened 2020-01-30
HOCS-1462 - Timeline is not ordered to milliseconds precision

hubert-raich:


MarkWrightBjss:
Build is currently failing.

hubert-raich:
Build fixed :) Broken test fixed with handy trick forcing the executorService to trigger actions on current thread (only when testing).

PR #298, opened 2020-02-20
HOCS-1444 deserialize the comma separate string into a list

lewismuirhead:


src/main/java/uk/gov/digital/ho/hocs/casework/api/dto/SearchRequest.java
```diff
@@ -17,6 +21,11 @@
     private String reference;
 
     @JsonProperty("caseType")
+    @JsonSetter
+    private void caseTypeDeserialize(String caseType) {
+        this.caseTypes = Arrays.asList(caseType.split(","));
```
MarkWrightBjss: Shockingly bad code. I expected much better from you.

lewismuirhead: learnt from the best ;-)

PR #306, opened 2020-03-25
HOCS-1640 : editing and deleting case notes API

MarkWrightBjss:
Editing and deleting case notes API


src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseNoteResource.java
```diff
@@ -46,4 +46,18 @@ public CaseNoteResource(CaseNoteService caseNoteService) {
         CaseNote caseNote= caseNoteService.createCaseNote(caseUUID, createCaseNoteRequest.getType(), createCaseNoteRequest.getText());
         return ResponseEntity.ok(caseNote.getUuid());
     }
+
+    @Authorised(accessLevel = AccessLevel.OWNER)
```
hubert-raich: Shouldn't the access level be the same as the one for createCaseNote()?

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseNoteService.java
```diff
@@ -58,4 +58,34 @@ public CaseNote createCaseNote(UUID caseUUID, String caseNoteType, String text)
...
         return caseNote;
     }
+
+    public CaseNote updateCaseNote(UUID caseNoteUUID, String caseNoteType, String text) {
+        log.debug("Updating CaseNote: {}", caseNoteUUID);
+        CaseNote caseNote = caseNoteRepository.findByUuid(caseNoteUUID);
+        if (caseNote != null){
+            String prevCaseNoteType = caseNote.getCaseNoteType();
+            String prevText = caseNote.getText();
+            caseNote.setCaseNoteType(caseNoteType);
```
hubert-raich: We need to keep the lastUpdatedBy information + lastUpdated timestamp. This should then be fetch so that frontend can use this info.

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseNoteResource.java
MarkWrightBjss: I down graded the Create endpoint so that the Task Force team can add notes.
The Task Force team should not be allowed to edit/delete case notes - that'll be more a manager role.

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseNoteService.java
MarkWrightBjss: Fixed.

MarkWrightBjss:
I've now made the changes to the Timeline items.

PR #309, opened 2020-04-02
HOCS-1581 : required by date for M-Ref cases

MarkWrightBjss:
Set different M-Ref deadline for whole Case lifetime, not per stage.
New endpoint to set deadline using different SLA (in days).
Calculation for Stage deadline includes Case deadline date for defaulting.
Merge to UKVI2.


hubert-raich:
changes look fine but the build failed:(

PR #310, opened 2020-04-09
UKVI2 HOCS-1589 : add correspondent

MarkWrightBjss:
New endpoint. Merge to UKVI2.

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/casework/api/CorrespondentService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/casework/api/CorrespondentService.java
@@ -66,8 +66,8 @@ public class CorrespondentService {
 
     Set<CorrespondentTypeDto> getCorrespondentTypes(UUID caseUUID){
         log.debug("Getting all Correspondent Types for Case: {}", caseUUID);
-        CaseData caseData = caseDataRepository.findByUuid(caseUUID);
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/digital/ho/hocs/casework/api/CorrespondentService.java
```diff
@@ -49,6 +64,15 @@ Correspondent getCorrespondent(UUID caseUUID, UUID correspondentUUID) {
         }
     }
 
+    Set<CorrespondentTypeDto> getCorrespondentTypes(UUID caseUUID){
+        log.debug("Getting all Correspondent Types for Case: {}", caseUUID);
+        CaseData caseData = caseDataRepository.findByUuid(caseUUID);
```
hubert-raich, line 69: We only need 'caseType' here. Let's create a method that retrieves case type for case uuid.

MarkWrightBjss: Do you mean add a method to the CaseData Repository that returns only the Type string, rather than the whole CaseData object?

hubert-raich: Yes, similar to this one 'String getCaseRef(UUID uuid);'.

PR #311, opened 2020-04-17
UKVI2 HOCS-1582 : correspondence cases are sent to correct workstack

MarkWrightBjss:
New endpoint to create the link value from the case data.
Merge to UKVI2.

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
@@ -12,12 +12,14 @@ import uk.gov.digital.ho.hocs.casework.api.dto.CaseDataType;
 import uk.gov.digital.ho.hocs.casework.api.dto.FieldDto;
 import uk.gov.digital.ho.hocs.casework.api.dto.GetStandardLineResponse;
 import uk.gov.digital.ho.hocs.casework.api.dto.TemplateDto;
+import uk.gov.digital.ho.hocs.casework.application.LogEvent;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
```diff
@@ -339,7 +341,31 @@ CaseSummary getCaseSummary(UUID caseUUID) {
...
+        // build the linkValue text string used to search the team link table by converting "text" key to the case's data value
+        String linkValue = null;
+        for (String text : texts){
+            String value = dataMap.getOrDefault(text, "");
+            if (!value.isEmpty()) {
+                if (linkValue != null) {
+                    linkValue += "_";
+                    linkValue += value;
+                } else {
+                    linkValue = value;
```
hubert-raich, line 358: Are we sure we want to reset the linkValue to blank string if the last text is not found? This would get rid of any previously allocated linkValues. I assume current logic (workflow) is built so that we always match on all text values, i'm not sure if this is always going to be the case going forward

MarkWrightBjss: If the last text isn't found then `if (!value.isEmpty()) {` means nothing happens. Line 358 only happens when (i) the last text _is_ found and, (ii) linkValue is null (hasn't been loaded with anything yet, i.e. first time round).

hubert-raich: Yes, thats right...sorry i swear it looked different last time I looked at it :D

PR #316, opened 2020-05-06
HOCS-1701 - DCU Custom Report Changes

hubert-raich:


src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataResource.java
```diff
@@ -134,4 +134,15 @@ ResponseEntity clearCachedTemplateForCaseType(@PathVariable String caseType) {
...
     }
+
+    @GetMapping(value = "/case/{caseUUID}/data/{variableName}")
+    ResponseEntity<String> getCaseDataValue(@PathVariable UUID caseUUID, @PathVariable String variableName) {
+        return ResponseEntity.ok(caseDataService.getCaseDataField(caseUUID, variableName));
+    }
+
+    @PutMapping(value = "/case/{caseUUID}/data/{variableName}")
+    ResponseEntity updateCaseDataValue(@PathVariable UUID caseUUID, @PathVariable String variableName, @RequestBody String value) {
+        caseDataService.updateCaseData(caseUUID, null, Map.of(variableName, value));
```
MarkWrightBjss, line 145: Will passing NULL into Stage UUID lose auditing, and reporting?

hubert-raich: I was worried about it as well. From my testing it looked like there was no impact at all. I'm not passing a stage uuid here as the update takes place in between 2 stages (on top level workflow). This is also why i added new endpoint that does not verify that a user is assigned to a case. Thorough testing of reporting mechanism will be performed in qa as part of this jira.

PR #323, opened 2020-05-26
HOCS-1751

Paulcr-ho:
Updating CI scripts with new tokens for the HOCS service separation.
We no longer have everything under CS. They are separated into namespaces.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -72,7 +72,7 @@ pipeline:
       - VERSION=build-${DRONE_BUILD_NUMBER}
       - KUBE_SERVER=https://kube-api-notprod.notprod.acp.homeoffice.gov.uk
     secrets:
-      - hocs_casework_dev
+      - hocs_casework_dev_cs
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -127,13 +130,46 @@ pipeline:
...
+      - ENVIRONMENT=prod
+      - KUBE_SERVER=https://kube-api-prod.prod.acp.homeoffice.gov.uk
+    secrets:
+      - hocs_casework_prod_cs
+    commands:
+      - cd kube-hocs-casework
+      - ./deploy.sh
+    when:
+      event: deployment
+      environment: prod
```
zuzak, line 163: This will mean when you deploy to prod you will deploy to both production environments; there won't be a way to deploy to just cs-prod or just wcs-prod.

Paulcr-ho: Well spotted, will modify for both qa and prod.

PR #324, opened 2020-05-27
Feature/hocs 1751 ci drone updates

Paulcr-ho:
Added the namespace filters for qa and prod.


zuzak:
I don't think this will do anything? `kube_namespace` is not a valid [step condition](https://0-8-0.docs.drone.io/step-conditions/).

PR #331, opened 2020-06-17
Feature/hocs 1671

parth-ho:


src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataResource.java
```diff
@@ -96,6 +96,16 @@ ResponseEntity updateCasePrimaryCorrespondent(@PathVariable UUID caseUUID, @Path
         return ResponseEntity.ok().build();
     }
 
+    @PutMapping(value = "/case/{caseUUID}/stage/{stageUUID}/updatePrimaryCorrespondent")
+    ResponseEntity updatePrimaryCorrespondent(
```
hubert-raich, line 100: I'm not sure this is needed. Coudn't we use the method above ? 'updateCasePrimaryCorrespondent'

hubert-raich: Checked with Parth, i thought the mapping would work fine between UUID and String, turns out it does not do so out of the box so I'm happy to accept this solution

PR #343, opened 2020-09-10
HOCS-1940-Allocations-Correspondents-MI

Paulcr-ho:
Added a rest endpoint for case reference lookups with the uuid.

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataResource.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataResource.java
@@ -185,4 +185,12 @@ class CaseDataResource {
         CaseData caseData = caseDataService.getCaseDataByReference(decodedRef);
         return ResponseEntity.ok(GetCaseResponse.from(caseData, true));
     }
+
+    @Authorised(accessLevel = AccessLevel.READ)
+    @GetMapping(value = "/case/reference/{caseUUID}")
+    ResponseEntity<GetCaseReferenceResponse> getCaseReference(@PathVariable UUID caseUUID) {
+        final String caseRef = caseDataService.getCaseRef(caseUUID);
+        return ResponseEntity.ok(GetCaseReferenceResponse.from(caseUUID, caseRef));
+    }
+
 }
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataResource.java
```diff
@@ -185,4 +185,12 @@ ResponseEntity updateCaseDataValue(@PathVariable UUID caseUUID, @PathVariable St
         CaseData caseData = caseDataService.getCaseDataByReference(decodedRef);
         return ResponseEntity.ok(GetCaseResponse.from(caseData, true));
     }
+
+    @Authorised(accessLevel = AccessLevel.READ)
```
hubert-raich, line 189: I don't think we need to restrict access to this method. It will just make it a tiny bit slower.

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataResource.java
```diff
@@ -185,4 +185,12 @@ ResponseEntity updateCaseDataValue(@PathVariable UUID caseUUID, @PathVariable St
...
+    @GetMapping(value = "/case/reference/{caseUUID}")
+    ResponseEntity<GetCaseReferenceResponse> getCaseReference(@PathVariable UUID caseUUID) {
```
hubert-raich, line 191: We should cache this method using spring cache. Also we should cache null responses (most caching annotations we have do exclude null result caching). This will help to minimise the non case uuid requests.

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataResource.java
```diff
@@ -185,4 +185,12 @@ ResponseEntity updateCaseDataValue(@PathVariable UUID caseUUID, @PathVariable St
...
+        final String caseRef = caseDataService.getCaseRef(caseUUID);
```
hubert-raich, line 192: This method does not return a reference for deleted cases. I guess we can update it or create a new method. It doesn't look like it is important in any usage before this one.

hubert-raich (changes requested):

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataResource.java
Paulcr-ho: Agreed

hubert-raich (approved):

PR #352, opened 2020-11-16
HOCS-1752: UKVI - take next case button

parth-ho:


src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
```diff
@@ -218,6 +219,35 @@ void updateStageUser(UUID caseUUID, UUID stageUUID, UUID newUserUUID) {
         return stages;
     }
 
+    Stage getUnassignedAndActiveStageByTeamUUID(UUID teamUUID, UUID userUUID) {
+        log.debug("Getting unassigned stage for user: {} in team {}", userUUID, teamUUID);
```
zuzak, line 223: Can we change this log line to make it clearer this is the "next unassigned case" button?
```suggestion
        log.debug("Getting next unassigned stage for user: {} in team {}", userUUID, teamUUID);
```

parth-ho: Yeah I did think about using either "case" or "stage", but the convention seems to be a mix of both and also we're technically retrieving the "stage" entity. The data model around this is also a bit muddled. But yeah I suppose for all practical purposes we're getting and assigning the "case". Will change the naming.

Paulcr-ho (approved):

PR #354, opened 2020-11-24
HOCS-2195: Correct indexes on somu item table

LiamMacP:
When creating the somu_item table an incorrect index was created 'idx_somu_item_uuid' that pointed to somu_item(case_uuid). This should have been pointing to somu_item(uuid).

There is also a redundant index in 'idx_somu_item_case_uuid' as a result of 'idx_somu_item_case_uuid_type'. This is because indexes work left to right in subsets. so this indexes case_uuid and case_uuid and type together.


Paulcr-ho:
Looks like the gitignore file was committed by mistake.

LiamMacP:
> Looks like the gitignore file was committed by mistake.

It wasn't a mistake, it is there to stop the staging for people that use Mac's as it's plaguing my unstaged files with .DS_Store files under each directory.

Paulcr-ho (approved):

PR #360, opened 2020-12-17
HOCS-2235: change somu item to upsert on no retrieval

LiamMacP:
At present when we try and get a somu item, if the item is not found then we currently throw an entity is not found exception. Whereas the new functionality means that a dummy somu item is created (that's classed as deleted) that will allow for the frontend to add and populate data based on the somu item uuid.

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/casework/api/SomuItemService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/casework/api/SomuItemService.java
@@ -45,29 +45,35 @@ public class SomuItemService {
             return somuItem;
         }
         else {
-            throw new ApplicationExceptions.EntityNotFoundException(String.format(NOT_FOUND_MESSAGE, somuTypeUuid), SOMU_ITEM_NOT_FOUND);
+            log.info("Could not find somu item for case {} for somu type {}, creating one. Event {}", caseUuid, somuTypeUuid, value(EVENT, SOMU_ITEM_RETRIEVED));
+            return addSomuItem(caseUuid, somuTypeUuid, null);
         }
     }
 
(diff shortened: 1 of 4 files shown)


Paulcr-ho (approved):

src/main/java/uk/gov/digital/ho/hocs/casework/api/SomuItemService.java
```diff
@@ -45,31 +45,37 @@ public SomuItem getSomuItem(UUID caseUuid, UUID somuTypeUuid) {
...
         else {
-            throw new ApplicationExceptions.EntityNotFoundException(String.format(NOT_FOUND_MESSAGE, somuTypeUuid), SOMU_ITEM_NOT_FOUND);
+            log.info("Could not find somu item for case {} for somu type {}, creating one. Event {}", caseUuid, somuTypeUuid, value(EVENT, SOMU_ITEM_RETRIEVED));
+            return addSomuItem(caseUuid, somuTypeUuid, null);
         }
     }
 
-    public SomuItem upsertSomuItem(UUID caseUUID, UUID somuTypeuuid, String data) {
-        log.debug("Upserting Somu Item of Type: {} for Case: {}", somuTypeuuid, caseUUID);
+    public SomuItem upsertSomuItem(UUID caseUuid, UUID somuTypeUuid, String data) {
```
MarkWrightBjss, line 53: I appreciate your effort to standardise the capitalisation here, but I think "UUID" is more common.

LiamMacP: As agreed, this will form part of the next PR in the series.

PR #362, opened 2021-01-04
HOCS-2254-Allow multiple value fields for each check field

agraham19:
HOCS-2254 adds additional uplift amounts.

These values need to be totalled up in Total Awarded and Total Paid.

Currently these totals are made of lists of fields to add and fields to subtract. Each field has a corresponding "check field", the value of which is used to determine whether to include the field in the total. As this check field is also used as the key to the map of fields to add, it is currently impossible to add two value fields with the same check field.

This change allows an array to be passed with each check field, whilst still accepting single values. I chose this solution as it allows existing projects to remain unaffected.

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/DataTotal.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/DataTotal.java
@@ -3,19 +3,24 @@ package uk.gov.digital.ho.hocs.casework.domain.model;
 import org.apache.commons.lang3.StringUtils;
 
 import java.math.BigDecimal;
+import java.util.List;
 import java.util.Map;
+import java.util.function.BiFunction;
 
 public class DataTotal {
 
-    public BigDecimal calculate(Map<String,String> dataMap, Map<String,String> addFields, Map<String,String> subFields){
-        BigDecimal total = BigDecimal.ZERO;
-        for(Map.Entry<String, String> field : addFields.entrySet()) {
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/DataTotal.java
```diff
@@ -3,19 +3,23 @@
...
-    public BigDecimal calculate(Map<String,String> dataMap, Map<String,String> addFields, Map<String,String> subFields){
-        BigDecimal total = BigDecimal.ZERO;
-        for(Map.Entry<String, String> field : addFields.entrySet()) {
-            total = total.add(parseCurrency(dataMap, field.getKey(), field.getValue()));
-        }
-        for(Map.Entry<String, String> field: subFields.entrySet()) {
-            total = total.subtract(parseCurrency(dataMap, field.getKey(), field.getValue()));
-        }
-        return total;
+    BiFunction<Map<String,String>, Map<String, List<String>>, BigDecimal> sumFields = (dataMap, fields) -> {
```
LiamMacP, line 12: Really like the use of BiFunction here, for some people though this might be a little complicated (maybe worth adding a quick comment here to explain a little.

LiamMacP (approved):

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/DataTotal.java
agraham19: I've added a little comment here

PR #368, opened 2021-01-11
HOCS-2055 Topic column on workstack team queue

robertdeniszczyc2:


src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/Stage.java
```diff
@@ -87,6 +87,10 @@
     @Column(name = "case_created", insertable = false, updatable = false)
     private LocalDateTime caseCreated;
 
+    @Getter
+    @Column(name = "text", insertable = false, updatable = false)
```
LiamMacP, line 91: Is text a good name for this? could we make this a little clearer with an alias?

src/main/resources/db/migration/postgresql/V1_8__2055_Topic_Column_On_Workstack_Team_Queue.sql
```diff
@@ -0,0 +1,12 @@
+CREATE OR REPLACE VIEW stage_data AS
+    SELECT c.reference AS case_reference, c.type AS case_type, c.data as data, c.created AS case_created, s.*, t.text
```
LiamMacP, line 2:
```suggestion
    SELECT c.reference AS case_reference, c.type AS case_type, c.data as data, c.created AS case_created, s.*, t.text AS case_assigned_topic
```

Then use this above?

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/Stage.java
robertdeniszczyc2: Updated with alias

src/main/resources/db/migration/postgresql/V1_10__2055_Topic_Column_On_Workstack_Team_Queue.sql
```diff
@@ -0,0 +1,16 @@
+DROP VIEW IF EXISTS stage_data;
```
parth-ho, line 1: Do we want to get rid of all exiting stage data?

robertdeniszczyc2: As I understand views they are virtual table/composition table created from "real tables", and don't hold the actual data, so wouldn't be removing real data, just the composition of the view? I added this in when I added the alias which was recommended by @LiamMacP , as without it the query wouldn't run. Happy to alter if there is a better way of achieving this though.

src/main/resources/db/migration/postgresql/V1_10__2055_Topic_Column_On_Workstack_Team_Queue.sql
```diff
@@ -0,0 +1,16 @@
...
+
+CREATE OR REPLACE VIEW stage_data AS
```
LiamMacP, line 3: Do we need to drop the views, the create or replace view will do this for you? views act as virtual tables and thus from my knowledge don't hold info.

src/main/resources/db/migration/postgresql/V1_10__2055_Topic_Column_On_Workstack_Team_Queue.sql
```diff
@@ -0,0 +1,16 @@
...
+    SELECT c.reference AS case_reference, c.type AS case_type, c.data as data, c.created AS case_created, s.*, t.text AS case_assigned_topic
+    FROM stage s
+            JOIN active_case c ON s.case_uuid = c.uuid
+            LEFT JOIN casework.topic t ON c.primary_topic_uuid = t.uuid;
+
+DROP VIEW IF EXISTS active_stage_data;
```
LiamMacP, line 9: Do we need to drop the views, the create or replace view will do this for you? views act as virtual tables and thus from my knowledge don't hold info.

src/main/resources/db/migration/postgresql/V1_10__2055_Topic_Column_On_Workstack_Team_Queue.sql
robertdeniszczyc2: I'll try it without

parth-ho: Yeah `stage_data` is a bit of a combination of data, but records from `stage_data` are retrieved from the frontend on specific actions. So I'm thinking if we we get rid of this view, actions for existing cases will break as it will no longer have the historical data it needs for some of the queries.

parth-ho: See `hocs-casework/src/main/java/uk/gov/digital/ho/hocs/casework/domain/repository/StageRepository.java` for examples of where we rely on data in `stage_data`

robertdeniszczyc2: So without it the application fails to run due to:
`2021-01-12 12:09:19.979  WARN 49767 --- [           main] ConfigServletWebServerApplicationContext : Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'flywayInitializer' defined in class path resource [org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration$FlywayConfiguration.class]: Invocation of init method failed; nested exception is org.flywaydb.core.internal.command.DbMigrate$FlywayMigrateException: 
Migration V1_10__2055_Topic_Column_On_Workstack_Team_Queue.sql failed
---------------------------------------------------------------------
SQL State  : 42P16
Error Code : 0
Message    : ERROR: cannot change name of view column "correspondents" to "case_assigned_topic"
Location   : db/migration/postgresql/V1_10__2055_Topic_Column_On_Workstack_Team_Queue.sql (/Users/robertdeniszczyc/repos/hocs-casework/build/resources/main/db/migration/postgresql/V1_10__2055_Topic_Column_On_Workstack_Team_Queue.sql)
Line       : 1
Statement  : CREATE OR REPLACE VIEW stage_data AS
    SELECT c.reference AS case_reference, c.type AS case_type, c.data as data, c.created AS case_created, s.*, t.text AS case_assigned_topic
    FROM stage s
            JOIN active_case c ON s.case_uuid = c.uuid
            LEFT JOIN casework.topic t ON c.primary_topic_uuid = t.uuid

2021-01-12 12:09:20.009 ERROR 49767 --- [           main] o.s.boot.SpringApplication               : Application run failed`

cc @LiamMacP as it mentions the correspondents stuff

.drone.yml
```diff
@@ -7,7 +7,7 @@ pipeline:
       - export DB_HOST='postgres'
       - export AWS_LOCAL_HOST='localstack'
       - export HOCS_DOCUMENT_SERVICE='http://documents:8080'
-      - ./gradlew build --no-daemon
+      - ./gradlew build --no-daemon --info
```
LiamMacP, line 10: This can be removed now as no longer required.

LiamMacP, line 10:
```suggestion
      - ./gradlew build --no-daemon
```

No longer required to have the `--info`

LiamMacP (changes requested):

.drone.yml
robertdeniszczyc2: Good spot, have removed

src/main/resources/db/migration/postgresql/V1_10__2055_Topic_Column_On_Workstack_Team_Queue.sql
agraham19: This has been removed.

agraham19 (approved):

LiamMacP (approved):
Changes made, passes builds, approved.

PR #369, opened 2021-01-12
HOCS-2301-"Unknown exception" exception only logs exception message

agraham19:
* add event and stacktrace to uncaught exceptions log output
* write tests for error handler

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/casework/domain/exception/RestResponseEntityExceptionHandler.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/casework/domain/exception/RestResponseEntityExceptionHandler.java
@@ -75,7 +75,7 @@ public class RestResponseEntityExceptionHandler {
 
     @ExceptionHandler(Exception.class)
     public ResponseEntity handle(Exception e) {
-        log.error("Exception: {}", e.getMessage(), value(EVENT, UNCAUGHT_EXCEPTION));
+        log.error("An error occurred", e, value(EVENT, UNCAUGHT_EXCEPTION));
         return new ResponseEntity<>(e.getMessage(), INTERNAL_SERVER_ERROR);
     }
 
--- a/src/test/java/uk/gov/digital/ho/hocs/casework/domain/exception/RestResponseEntityExceptionHandlerTest.java
+++ b/src/test/java/uk/gov/digital/ho/hocs/casework/domain/exception/RestResponseEntityExceptionHandlerTest.java
@@ -1,0 +1,181 @@
+package uk.gov.digital.ho.hocs.casework.domain.exception;
+
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.core.read.ListAppender;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import ch.qos.logback.classic.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.ResponseEntity;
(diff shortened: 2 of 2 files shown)


src/main/java/uk/gov/digital/ho/hocs/casework/domain/exception/RestResponseEntityExceptionHandler.java
```diff
@@ -75,7 +79,10 @@ public ResponseEntity handle(HttpMessageNotReadableException e) {
 
     @ExceptionHandler(Exception.class)
     public ResponseEntity handle(Exception e) {
-        log.error("Exception: {}", e.getMessage(), value(EVENT, UNCAUGHT_EXCEPTION));
+        Writer stackTraceWriter = new StringWriter();
+        PrintWriter printWriter = new PrintWriter(stackTraceWriter);
+        e.printStackTrace(printWriter);
```
zuzak, line 84: Does this get output as JSON? If not, it'll be a pain to dig out of Kibana

src/main/java/uk/gov/digital/ho/hocs/casework/domain/exception/RestResponseEntityExceptionHandler.java
```diff
@@ -75,7 +79,10 @@ public ResponseEntity handle(HttpMessageNotReadableException e) {
 
     @ExceptionHandler(Exception.class)
     public ResponseEntity handle(Exception e) {
-        log.error("Exception: {}", e.getMessage(), value(EVENT, UNCAUGHT_EXCEPTION));
+        Writer stackTraceWriter = new StringWriter();
+        PrintWriter printWriter = new PrintWriter(stackTraceWriter);
```
parth-ho, line 83: Could we not do something like `e.getStackTrace().toString()` instead? Not sure if we need to do this as I think the Exception object already has everything we need.

agraham19: Unfortunately this just prints [Ljava.lang.StackTraceElement;@... I've done a quick Google, but there doesn't seem to be a simpler way to stringify the stacktrace.

src/main/java/uk/gov/digital/ho/hocs/casework/domain/exception/RestResponseEntityExceptionHandler.java
agraham19: The logback configuration should output the logs as json based on the profile used in dev.

src/main/java/uk/gov/digital/ho/hocs/casework/domain/exception/RestResponseEntityExceptionHandler.java
```diff
@@ -75,7 +75,7 @@ public ResponseEntity handle(HttpMessageNotReadableException e) {
 
     @ExceptionHandler(Exception.class)
     public ResponseEntity handle(Exception e) {
-        log.error("Exception: {}", e.getMessage(), value(EVENT, UNCAUGHT_EXCEPTION));
+        log.error("An error occurred", e);
```
parth-ho, line 78: We still need `value(EVENT, UNCAUGHT_EXCEPTION)` as that seems to be the way be define logs using this key value pair.

agraham19: I have readded this.

parth-ho (approved):

PR #373, opened 2021-01-29
HOCS-2027: MPAM global permissions

parth-ho:


philbarton (changes requested):
There are no changes to SecurityIntegrationTest in this project, yet there are in workflow. Should there be?

philbarton (approved):
Looks good. There is a lot here, so thorough integration testing will be required.

PR #374, opened 2021-02-08
HOCS-2027: Add empty check when getting admin case types.

parth-ho:


agraham19 (approved):

agraham19 (changes requested):
Build is failing

zuzak:
Can we add a test to catch whatever caused the build to silently break?

agraham19 (approved):

PR #377, opened 2021-03-16
HOCS-2555 - New endpoint to return active stage for a given case.

philbarton:
The case-creator service needs a way to find out the current active stage for a case.
This endpoint was added to allow that query.

It does return a Set but I believe that there can only be one active stage for a case. (I think the Set is a JPA limitation)

The client side code checks the Set length and if there is more than one it will raise an exception.

### DIFF
--- a/src/test/java/uk/gov/digital/ho/hocs/casework/api/StageResourceTest.java
+++ b/src/test/java/uk/gov/digital/ho/hocs/casework/api/StageResourceTest.java
@@ -148,7 +148,7 @@ public class StageResourceTest {
 
         when(stageService.getActiveStagesByCaseUUID(caseUUID)).thenReturn(stages);
 
-        ResponseEntity<GetStagesResponse> response =  stageResource.getActiveStagesByCase(caseUUID);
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/digital/ho/hocs/casework/api/StageResourceTest.java
```diff
@@ -141,6 +141,23 @@ public void shouldAllocateStageNull() {
...
     }
 
+    @Test
+    public void shouldGetActiveStagesCaseUUID() {
+        Set<Stage> stages = new HashSet<>();
+        UUID caseUUID = UUID.randomUUID();
+
+        when(stageService.getActiveStagesByCaseUUID(caseUUID)).thenReturn(stages);
+
+        ResponseEntity<GetStagesResponse> response =  stageResource.getActiveStagesByCase(caseUUID);
```
robertdeniszczyc2, line 151: Remove extra space after = and before stageResource
```suggestion
        ResponseEntity<GetStagesResponse> response = stageResource.getActiveStagesByCase(caseUUID);
```

PR #378, opened 2021-03-17
HOCS-2640: migrate from drone 0.8 to 1

LiamMacP:
This change implements the necessary changes to use Drone V1 as V0.8
is going to be deprecated. Changes include:
- Split build and deploy into two pipelines
- Removed pull request checks
- Changed to push to build_XXX instead of build-XXX
- Added commit sha tag
- Switched to using KUBE_TOKEN as defined in Drone rather than the kube 
repo
- Added acyclic dependency graph for asynchronous build steps
- Made a QA release always release to both QA environments
consolidated build steps where possible

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,215 +1,260 @@
-pipeline:
-
-  build-project:
-    image: quay.io/ukhomeofficedigital/openjdk11:v11.0.5_10
-    commands:
-      - export SPRING_PROFILES_ACTIVE='development, local'
-      - export DB_HOST='postgres'
-      - export AWS_LOCAL_HOST='localstack'
-      - export HOCS_DOCUMENT_SERVICE='http://documents:8080'
-      - ./gradlew build --no-daemon
-    when:
-      event: [push, pull_request, tag]
-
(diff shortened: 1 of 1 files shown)


.drone-v1.yml
```diff
@@ -0,0 +1,253 @@
...
+steps:
+  - name: build project
+    image: quay.io/ukhomeofficedigital/openjdk11:v11.0.5_10
+    environment:
+      SPRING_PROFILES_ACTIVE: "development, local"
+      HOCS_DOCUMENT_SERVICE: "http://documents:8080"
+      DB_HOST: "postgres"
+      AWS_LOCAL_HOST: "localstack"
+    commands:
+      - ./gradlew build --no-daemon
```
zuzak, line 22: Let's split this into two steps: `./gradlew assemble --no-daemon` and `./gradlew check --no-daemon` (`gradlew build` is essentially an alias for these two things anyway).

LiamMacP: This has been completed

zuzak (approved):

PR #380, opened 2021-04-01
HOCS-2457 add Spring config to allow Swagger service on casework

abarker66:
This ticket adds a Swagger service to casework giving us access to all the exposed api endpoints published by the casework service.
The Swagger page is available at http://localhost:8082/swagger-ui.html

### DIFF
--- a/bin/.gitignore
+++ b/bin/.gitignore
@@ -1,0 +1,3 @@
+/main/
+/test/
+/default/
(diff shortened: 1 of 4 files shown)


robertdeniszczyc2 (approved):
Looks good to me

robertdeniszczyc2 (changes requested):
As with info service as this uses identical implementation, has the acceptance criteria around being available on dev only being covered?

Thanks,
Rob

abarker66:
Spring profile switching is now added. The profile is 'swagger' and can be added as an env variable or as a -D JVM command line option. So without the 'swagger' profile there will be no swagger endpoints available.

robertdeniszczyc2 (approved):

PR #392, opened 2021-06-24
HOCS-3103 provide somu data alongside case data

gardnersam:


src/main/resources/db/migration/postgresql/V1_13__3103_AddSomuDatatoCase.sql
```diff
@@ -0,0 +1,20 @@
+--ALTER TABLE case_data
```
philbarton, line 1: Is this comment needed?

philbarton (commented):
This looks ok to me but might be worth asking Liam to have a quick look.

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
```diff
@@ -214,11 +218,26 @@ void updateStageUser(UUID caseUUID, UUID stageUUID, UUID newUserUUID) {
     Set<Stage> getActiveStagesByTeamUUID(UUID teamUUID) {
         log.debug("Getting Active Stages for Team: {}", teamUUID);
         Set<Stage> stages = stageRepository.findAllActiveByTeamUUID(teamUUID);
+        addSomuData(stages);
```
LiamMacP, line 221: I'm a little confused why on a get we are adding data, so probably worth some comments?

src/main/resources/db/migration/postgresql/V1_13__3103_AddSomuDatatoCase.sql
```diff
@@ -0,0 +1,20 @@
...
+         LEFT JOIN casework.correspondents_json_by_case cs ON s.case_uuid = cs.case_uuid
+         LEFT JOIN casework.topic t ON c.primary_topic_uuid = t.uuid
+WHERE s.team_uuid NOTNULL;
+
+CREATE OR REPLACE VIEW casework.stage_data AS
+SELECT c.reference AS case_reference, c.type AS case_type, c.data as data, c.created AS case_created, s.id, s.uuid, s.created, s.type, s.deadline, s.transition_note_uuid, s.case_uuid, s.team_uuid, s.user_uuid, s.deadline_warning, correspondents, t.text AS case_assigned_topic, s.somu
+FROM casework.stage s
+         JOIN casework.active_case c ON s.case_uuid = c.uuid
+         LEFT JOIN casework.correspondents_json_by_case cs ON s.case_uuid = cs.case_uuid
+         LEFT JOIN casework.topic t ON c.primary_topic_uuid = t.uuid;
```
LiamMacP, line 20: Although it'll not matter for this persay, for git diffs to work nicely, it should have a new line at the end.

src/main/resources/db/migration/postgresql/V1_13__3103_AddSomuDatatoCase.sql
gardnersam: Nope, removed

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
gardnersam: I've added a comment but to clarify here aswell the front end calls this for its information when loading the team dashboards and it returns information from the case_data table, the data field in that table contains unreliable data for the case contributions. So to add contribution statuses I needed the reliable data from the somu_item table so I've added it to the data that is sent.

src/main/resources/db/migration/postgresql/V1_13__3103_AddSomuDatatoCase.sql
gardnersam: Fair point I've added one in

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
LiamMacP: Awesome, thanks for explaining 👍🏽

LiamMacP (approved):

PR #394, opened 2021-06-29
HOCS-3299: Add endpoint for active stage by Case UUID

ARobinson26:
- Added a new endpoint the allows you to get the active stage for a case by it's UUID.
- Added a new method in StageService that ensures only one stage is returned.

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
@@ -211,6 +211,20 @@ public class StageService {
         return stageRepository.findAllActiveByCaseUUID(caseUUID);
     }
 
+    Optional<Stage> getActiveStageByCaseUUID(UUID caseUUID){
+        log.debug("Getting Active Stage for Case: {}", caseUUID);
+        Set<Stage> stages = stageRepository.findAllActiveByCaseUUID(caseUUID);
+        switch (stages.size()){
+            case 0:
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
```diff
@@ -211,6 +211,20 @@ void updateStageUser(UUID caseUUID, UUID stageUUID, UUID newUserUUID) {
         return stageRepository.findAllActiveByCaseUUID(caseUUID);
     }
 
+    Stage getActiveStageByCaseUUID(UUID caseUUID){
```
SeanReedHomeOffice, line 214: It would be better to return an Optional<Stage> from this function.

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
```diff
@@ -211,6 +211,20 @@ void updateStageUser(UUID caseUUID, UUID stageUUID, UUID newUserUUID) {
...
+        log.debug("Getting Active Stages for Case: {}", caseUUID);
```
SeanReedHomeOffice, line 215: Debug message needs fixing. Stage without the 's'.

SeanReedHomeOffice (changes requested):

nickmcmahon01:
Closing PR because it's over 2 months old.

PR #395, opened 2021-07-05
HOCS-3087 Updating dockerfile and adding health check file

gardnersam:


scripts/HealthCheck.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Class used for container health check by docker compose
+ */
+public class HealthCheck {
+
+    public static void main(String[] args) throws IOException, InterruptedException {
+        var client = HttpClient.newHttpClient();
+        var request = HttpRequest.newBuilder()
+                .uri(URI.create("http://localhost:8080/actuator/health"))
```
LiamMacP, line 17: Could we use `server.port` to make this more resilient? https://github.com/UKHomeOffice/hocs-casework/blob/c6f7a1ba51a73b716ea057a67855f2d16142349d/src/main/resources/application.properties#L5

gardnersam: This file now uses a command line argument to set the end point to check so it is fully customisable

philbarton (approved):
Looks good to me

PR #399, opened 2021-07-08
HOCS-2822: add csv case data updater script

LiamMacP:
At present to update cases within the system we have to manually 
create the url requests to be able to update the data or change the 
live data through the frontend. This is easily error prone and 
requires a lot of time to achieve. To improve this mechanism this 
change adds a scripts that takes an absolute csv filepath and 
updates the data based on whats passed in. A new dry run option 
has been added whereby an `update` (case insensitive) second arg 
should be provided to actually send the requests. To make it clear 
what each of the updates are going to do, the script also logs each 
request and the value that it is updating. If the request is successful 
we now also log that. 
As the script is embedded within the `hocs-casework` image, we 
can call the service directly through localhost and use the injected 
SERVER_PORT environment variable.

### DIFF
--- a/scripts/CsvCaseDataUpdater.java
+++ b/scripts/CsvCaseDataUpdater.java
@@ -1,0 +1,297 @@
+package uk.gov.digital.ho.hocs;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.util.*;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Takes a CSV file and is used to update the case values for specific cases.
+ *
+ * Format of the csvs must be as follows: {caseUUID, [fields]+}, with the first
+ * row being the column headers.
+ *
+ * Fields that hold no value within the CSV are ignored when it comes to updating.
+ *
+ * To use this file you first need to add the CSV file to the K8 hocs-casework
+ * container through {@code kubectl cp [CSVFilePath] [PodName]:. -c hocs-casework}.
+ * You can then invoke this java file through the containers command line by running
+ * {@code java CsvCaseDataUpdater.java [AbsoluteCSVFilePath]}.
+ */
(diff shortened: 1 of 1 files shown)


scripts/CsvCaseDataUpdater.java
```diff
@@ -0,0 +1,253 @@
...
+
+    /*
+     * We inject this variable through deployment, so that the spring boot application knows what to run on.
+     * Default to the spring boot 8080, incase it's not present.
+     */
+    private static final String SERVER_PORT =
+            getEnvironmentVariableOrDefault("SERVER_PORT", "8080");
+    private static final String BASE_URL_FORMAT = "http://localhost:" + SERVER_PORT + "/case/%s/data/%s";
+
+    public static void main(String[] args) throws IOException {
```
philbarton, line 37: Could this have a dryrun option?
And log the requests that would be made.

scripts/CsvCaseDataUpdater.java
```diff
@@ -0,0 +1,253 @@
...
+    private static void updateCases(List<CaseData> allCaseData) {
+        final HttpClient client = HttpClient.newHttpClient();
+
+        for (CaseData caseData: allCaseData) {
+            final List<HttpRequest> requests = generateUpdateRequestForCase(caseData);
+
+            // Loop round each request and send to the casework service
+            requests.forEach(request -> {
+                try {
+                    final HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
```
philbarton, line 174: Could this log the changes made?
I'm not sure if possible but somehow log to Kibana?

scripts/CsvCaseDataUpdater.java
LiamMacP: Updated to accept a case-insentive 2nd param as `update`. Without this it defaults into dry-run mode so that the user cannot accidently send data.

scripts/CsvCaseDataUpdater.java
LiamMacP: Added logging into the program, for all of the requests and also the success messages. As we throw an exception on a invalid request/ non-200 result, that already acts as logging there.

philbarton (approved):
lgtm

PR #407, opened 2021-07-22
HOCS-3421: increase hikari connection pool size

LiamMacP:
During a Live Service incident we saw that there was occurences
whereby the default connection pool size of 10 were all being used
up. This was causing the pods to crash and fail the health check.
To remediate from this, the default connection pool size has been
increased to 20 while we investigate what the correct number for
the system should be.


philbarton (approved):
lgtm

PR #408, opened 2021-07-22
HOCS-3311 Add tags onto stages in workstacks

robertdeniszczyc2:
Adds tags field to stages, used on the frontend to indicate particular statuses about a case clearly on a workstack view.

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
@@ -46,6 +46,7 @@ public class StageService {
     private final CaseDataService caseDataService;
     private final StagePriorityCalculator stagePriorityCalculator;
     private final DaysElapsedCalculator daysElapsedCalculator;
+    private final StageTagsDecorator stageTagsDecorator;
     private final CaseNoteService caseNoteService;
     private final SomuItemService somuItemService;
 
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
```diff
@@ -218,11 +220,14 @@ void updateStageUser(UUID caseUUID, UUID stageUUID, UUID newUserUUID) {
     Set<Stage> getActiveStagesByTeamUUID(UUID teamUUID) {
         log.debug("Getting Active Stages for Team: {}", teamUUID);
         Set<Stage> stages = stageRepository.findAllActiveByTeamUUID(teamUUID);
+
         for (Stage stage : stages) {
             addSomuData(stage); //List of case contributions from somu table required to show contributions status on teams dashboards
             updatePriority(stage);
             updateDaysElapsed(stage);
+            decorateTags(stage);
```
gardnersam, line 228: The somu data line has just been commented out, assume this could cause simular issues of data bloat?

robertdeniszczyc2: Ah of course, I'll need to bring this branch up to date to get the change. The concern with the somu one was for cases with multiple contributions the somu data payload could be very large, here the tags we're adding are just a small array of strings, and currently just one string "HS" or an empty array so relatively small payload change.

philbarton (approved):
LGTM

PR #412, opened 2021-07-30
HOCS-3013: add dashboard endpoint

LiamMacP:
At present when the frontend makes a call to load the dashboard,
the `/stage` endpoint is hit. This call returns more information
then is actually required to show the dashboard. As such, this
change adds a `/dashboard` endpoint that returns only the
required information to show the dashboard. 

As the information required is aggregated from a number of
different tables, splitting these out to increase performance was
ideal. This solution therefore queries each of the key pieces of
information separately and zips them together to return
information for each stage/team. As a result of these queries not
returning a entity, we have to run native queries and reflect the
values into a new model Statistic.

---

### Performance Tests

When performance testing the old code, it started to slowdown 
exponentially at around 49 users, and shortly after 90 second 
timeouts were reported.

Screenshot 2021-07-29 at 16 37 07

This new code was tested on the same environment and managed
to achieve an average of 800ms responses with 200 users. 

DashboardRequest200Users

Although, what should be noted that these tests were done against
a single `hocs-casework` pod and load balancing would show better
results with more users.


src/main/java/uk/gov/digital/ho/hocs/casework/api/DashboardService.java
```diff
@@ -0,0 +1,68 @@
...
+        this.requestData = requestData;
+        this.statisticRepository = statisticRepository;
+        this.userPermissionsService = userPermissionsService;
+    }
+
+    Map<UUID, Map<String, Integer>> getDashboard() {
+        log.debug("Getting dashboard statistics for user");
+        Set<UUID> allUserteams = userPermissionsService.getUserTeams();
+        if (allUserteams.isEmpty()) {
+            log.warn("No teams - Returning 0 Stages", value(EVENT, TEAMS_STAGE_LIST_EMPTY));
```
zuzak, line 41: Can we log the relevant user (unless we already do?)

src/test/java/uk/gov/digital/ho/hocs/casework/api/DashboardResourceTest.java
```diff
@@ -0,0 +1,51 @@
...
+    @Test
+    public void getDashboard_returnsEmptyMap_returnEmptyObject() {
+        when(dashboardService.getDashboard()).thenReturn(Collections.emptyMap());
+
+        var response = dashboardResource.getDashboard();
+
+        assertThat(response).isNotNull();
+        assertThat(response.getStatusCode()).isEqualTo(HttpStatus.OK);
+    }
+}
```
zuzak, line 51: newline

src/main/java/uk/gov/digital/ho/hocs/casework/api/DashboardService.java
LiamMacP: Comes through to Kibana through the `X-Auth-UserId` header.

src/test/java/uk/gov/digital/ho/hocs/casework/api/DashboardResourceTest.java
LiamMacP: Fixed 🙂

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/Statistic.java
```diff
@@ -0,0 +1,30 @@
+package uk.gov.digital.ho.hocs.casework.domain.model;
+
+import lombok.AllArgsConstructor;
+import lombok.NoArgsConstructor;
+
+import java.util.UUID;
+
+@NoArgsConstructor
+@AllArgsConstructor
+public class Statistic {
```
philbarton, line 10: When I saw the word statistic I got  a bit confused, would DashboardSummary be  a better name?

philbarton (changes requested):
Looks good, and works locally.
The name Statistic threw me though!

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/Statistic.java
LiamMacP: The idea is that this could be used for other purposes, and is a statistic from the database, unsure that `DashboardSummary` would be a great name. In theory it would be good to get this reused for the actual workstacks. Happy to rename, just unsure what a better name would be.

philbarton: Summary?
Aggregate?

philbarton (approved):

PR #413, opened 2021-08-02
HOCS-3013: change query to filter out unworkable

LiamMacP:
Due to a change requested by WCS, a new Unworkable field has
been added to WCS case data. As they don't want these cases
showing in the dashboard statistics, we need to filter these out.


fewtrees-ho (approved):

philbarton (approved):
NIce work, cheers Liam

PR #414, opened 2021-08-09
HOCS-3525 - Case Timeline not updated when correspondent updated

agraham19:
* Add CORRESPONDENT_UPDATED to TIMELINE_EVENTS


gary-newbolt (approved):
lgtm

PR #420, opened 2021-08-19
HOCS-3645: add correspondent type display name

LiamMacP:
Add a decorator that will take a list of CorrespondentTypeDto and
Correspondents and populate the `typeDisplayName`field of the
Correspondent if the value is there.

This new decorator is then called within the CorrespondentService
when returning Correspondents.

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/casework/api/utils/CorrespondentTypeNameDecorator.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/casework/api/utils/CorrespondentTypeNameDecorator.java
@@ -1,0 +1,41 @@
+package uk.gov.digital.ho.hocs.casework.api.utils;
+
+import org.springframework.stereotype.Service;
+import uk.gov.digital.ho.hocs.casework.api.dto.CorrespondentTypeDto;
+import uk.gov.digital.ho.hocs.casework.domain.model.BaseCorrespondent;
+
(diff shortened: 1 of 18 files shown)


src/main/java/uk/gov/digital/ho/hocs/casework/api/utils/CorrespondentTypeNameDecorator.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public <T extends BaseCorrespondent> T addCorrespondentTypeName(Set<CorrespondentTypeDto> correspondentTypes,
+                                                                    T correspondent) {
+        if (correspondentTypes == null) {
+            return correspondent;
+        }
+
+        Optional<CorrespondentTypeDto> correspondentType = correspondentTypes.stream()
+                .filter(correspondentTypeDto -> Objects.equals(correspondentTypeDto.getType(), correspondent.getCorrespondentType()))
+                .findFirst();
```
gardnersam, line 33: How do we know if .findFirst() gives us the right thing?

LiamMacP: The correspondent type has a unique constraint on it in the db, there will only ever be one match.

gardnersam (approved):
Looks good

PR #421, opened 2021-08-25
HOC-3670: stage user endpoint

LiamMacP:
Add an endpoint that returns all the active stages that the user
should have access too. This endpoint calls a native query that first
gets the stages that the user is assigned too, and then cross-references
with the teams the user has access to. As we have access to the
case data at this stage, this change also filters on `Unworkable`.

On QAX when timing the current `\stage` endpoint vs this new one with 
50 users hitting the system at the same time, the response time by the
service was as follows:

`Class: uk.gov.digital.ho.hocs.casework.api.StageResource. Method: getActiveStages. Execution time: 9405ms`
`Class: uk.gov.digital.ho.hocs.casework.api.StageResource. Method: getActiveStagesForUser. Execution time: 88ms`


gardnersam (approved):
Looks good, nice improvements to surrounding code also

PR #423, opened 2021-09-06
Feature/hocs 3556 retrieve workstack next case type column

fewtrees-ho:
Casework Stage 2 implementation

https://collaboration.homeoffice.gov.uk/jira/browse/HOCS-3528


gardnersam (approved):
Looks good

PR #424, opened 2021-09-07
HOCS-3718 Extend Extraction Capability to Extract data from case_deadline_extension table

agraham19:
HOCS-3718 Extend Extraction Capability to Extract data from case_deadline_extension table
    
    * Send extract updates to audit with their own payload


gary-newbolt (commented):
lgtm - haven't approved yet until the Audit is fixed and I've given it the run through.

gary-newbolt (approved):
all good thanks

PR #428, opened 2021-09-14
HOCS-3787: Change failed response handling for CSVCaseDataUpdater

LiamMacP:
Currently, when we get a non-200 code from a request the rest of the
urls don't get requested. This can lead to partial updates which aren't
ideal. This change logs out a severe error when this occurs instead of
terminating the program.


gardnersam (approved):
Looks good

PR #429, opened 2021-09-15
HOCS-3774 adding contribution due date to data passed to mycases

gardnersam:
Problem was caused by the updateContributions method not being called on the stage objects in the updateStages() method. I decided to not update updateStages() as this is called by the stage endpoint which famously should not have more data added to it.


LiamMacP (approved):
Looks good to me

PR #431, opened 2021-09-16
Separate active_case and case_data entities

agraham19:
Separate active_case and case_data entities to fix Case Extensions


abarker66 (approved):
Nice work Andrew, the case data repo actually acts on the case data table now and the active case view has its own repo class.

PR #434, opened 2021-09-22
HOCS-3831: Fix missing Stage 2 Case Data

ARobinson26:
Three fields were missing when escalating a complaints case from stage 1 to stage 2. This adds the three fields to the copy class.


gardnersam (approved):
Looks good

PR #436, opened 2021-09-23
HOCS-3852 Make contribution processing handle leading zeros

gardnersam:
Adds a datetimeformatter to let LocalDate.parse() handle leading zeros and also a lack of zeros.


LiamMacP (approved):
Looks spot on to me!

Cool use of `thenCallRealMethod`...

PR #437, opened 2021-09-23
HOCS-3853 Decorate resposes to /stage calls with contribution informa…

gardnersam:
…tion


LiamMacP (approved):
Although this solution will need re-looking at in the long term, adding this follows the previous pattern.

PR #439, opened 2021-09-23
Fix/hocs 3855 fix active case query

fewtrees-ho:
https://collaboration.homeoffice.gov.uk/jira/browse/HOCS-3855


gardnersam (approved):
Timings look good, test coverage solid

PR #440, opened 2021-09-28
Feature/hocs 3869 noisy audit on workstack view

gardnersam:


nickmcmahon01 (approved):
LGTM

nickmcmahon01:
I'm not a huge fan of an optional audit trail in the method signature but not sure what else we can do without unpicking why somu items have their own endpoint and also need to be served as part of the stage data. Do we audit when we include a Correspondent as part of the stage data? What makes Somu special in this regard or are we just handling the two inconsistently?

LiamMacP:
> I'm not a huge fan of an optional audit trail in the method signature but not sure what else we can do without unpicking why somu items have their own endpoint and also need to be served as part of the stage data. Do we audit when we include a Correspondent as part of the stage data? What makes Somu special in this regard or are we just handling the two inconsistently?

The purpose of the somu endpoints are that they are their own 'thing'. The reason it's now linked to the stage is due to being shown in some workspaces as a column. I don't actually feel like we need this audit event as the user isn't actually viewing somu items. I don't believe we audit the correspondents. We should really go back and audit our own audit logs... 🙂

PR #449, opened 2021-10-08
HOCS-3858: get stage team through projection

LiamMacP:
When trying to get the team uuid currently this queries the `stage_data`
 view that returns not only more information than required but also
 computes every join. As the required data is immediately
 available on the `stage` table, this can easily be retrieved through a
 projection, saving the computation time of the view.

During testing on QA with around 70k current stages, the old query returned
in 60ms (without load) whilst the new one was around 1ms (without load).

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
@@ -105,7 +105,7 @@ public class StageService {
         Stage.StageTeamUuid stageTeamUuid = stageRepository.findTeamUuidByCaseUuidAndStageUuid(caseUUID, stageUUID);
 
         if (stageTeamUuid == null) {
-            log.info("No team exists is linked to stage: {} and case: {}", stageUUID, caseUUID);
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
```diff
@@ -102,9 +102,16 @@ public UUID getStageUser(UUID caseUUID, UUID stageUUID) {
 
     public UUID getStageTeam(UUID caseUUID, UUID stageUUID) {
         log.debug("Getting Team for Stage: {}", stageUUID);
-        Stage stage = getStage(caseUUID, stageUUID);
-        log.debug("Got Team: {} for Stage: {}", stage.getTeamUUID(), stageUUID);
-        return stage.getTeamUUID();
+        Stage.StageTeamUuid stageTeamUuid = stageRepository.findTeamUuidByCaseUuidAndStageUuid(caseUUID, stageUUID);
+
+        if (stageTeamUuid == null) {
+            log.info("No team exists is linked to stage: {} and case: {}", stageUUID, caseUUID);
```
fewtrees-ho, line 108: should this be a `warn`ing instead of `info`?

PR #452, opened 2021-10-14
Hocs 3779 fix foi topic showing as UUID

LaurieBennett:


src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
```diff
@@ -461,10 +462,18 @@ void updatePrimaryCorrespondent(UUID caseUUID, UUID stageUUID, UUID primaryCorre
...
     }
 
-    void updatePrimaryTopic(UUID caseUUID, UUID stageUUID, UUID primaryTopicUUID) {
+    void updatePrimaryTopic(UUID caseUUID, UUID stageUUID, UUID primaryTopicUUID, UUID textUUID) {
         log.debug("Updating Primary Topic for Case: {} Topic: {}", caseUUID, primaryTopicUUID);
         CaseData caseData = getCaseData(caseUUID);
         caseData.setPrimaryTopicUUID(primaryTopicUUID);
+        // If we have been passed a text UUID we should change the data to replace it in data with the topic UUID
+        // to allow
+        if (textUUID != null){
```
agraham19, line 471: I think rather than adding textUUID and checking for null on the existing method, we should create a new method that takes a textUUID. That way the original methods don't need to pass null.

PR #453, opened 2021-10-18
HOCS-3979: contribution processor improvements

LiamMacP:
Presently, when a workstack is loaded, a call is made for each 'stage'
is part of that workstack to retrieve the contributions. This is causing
potentially thousands of calls and taking more resources than necessary
This change requests all the contributions for all the stages at once,
leading to only one hit on the database.

A side effect of this is we can revert the auditing code that was
applied previously.


gardnersam (approved):
With the changes from the previous PR merged in this looks good

PR #463, opened 2021-11-03
HOCS-4073: use multi-stage docker file

nickmcmahon01:
This commit moves us to a multi-stage docker build with a layered jar instead of a fat jar. The objective is to reduce the springboot startup time in k8s environments, a side effect is more efficent docker layer caching for developer flow.

This change in isolation reduces the Startup time in k8s from 
Started HocsCaseApplication in 24.14 seconds (JVM running for 25.76)
to
Started HocsCaseApplication in 10.893 seconds (JVM running for 11.479)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/ukhomeofficedigital/alpine:v3.13 as builder
+FROM quay.io/ukhomeofficedigital/alpine:v3.14 as builder
 
 USER root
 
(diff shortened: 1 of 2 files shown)


zuzak:
Please amend your commit to have a brief subject line of fewer than 50 characters, followed by your longer explanation. See https://chris.beams.io/posts/git-commit/#limit-50.

LiamMacP (approved):
Looks good to me, I'd say squash this and put a useful commit message though...

PR #464, opened 2021-11-03
HOCS-4073 Use Spring-context-indexer

nickmcmahon01:
This commit uses spring-context-indexer to reduce autowiring startup time 

This change in isolation reduces the Startup time in k8s from
Started HocsCaseApplication in 24.14 seconds (JVM running for 25.76)
to
Started HocsCaseApplication in 10.893 seconds (JVM running for 11.479)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,39 @@
-FROM quay.io/ukhomeofficedigital/alpine:v3.13
-
-ENV USER user_hocs_casework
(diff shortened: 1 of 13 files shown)


nickmcmahon01:
There was a bug with  swagger 2.9.2 and Spring context indexer, this has been resolved in later versions.

LiamMacP (approved):

PR #471, opened 2021-11-10
HOCS-4058: fix display of appeals in summary

gary-newbolt:


abarker66 (approved):
All good. Are we sure that DECS is ready for fluent interfaces on its model classes (smile).

PR #473, opened 2021-11-12
HOCS-3871 - Record external interest in the case

agraham19:
* HOCS-3871 - Record external interest in the case


gary-newbolt (changes requested):
Apologies... I've merged my work on casework to split the api that will effect this. Loads of conflicts. Let me know if you want some help it should make the scope of the changes for this less though

PR #475, opened 2021-11-16
Fix/hocs 3972 fix extension appeals extracts

gary-newbolt:


abarker66 (approved):
LGTM. One thing (a general comment) is we have across the services a mix of persistence packages, spring data (JPA) and javax.persistence - I assume tit just evolved this way, there is no agreed approach. The Spring Data / JPA2 approach has built in Audit functionality - just annotate a field with @CreatedDate @UpdatedDate, etc - maybe I should let Rob S know :-)

PR #476, opened 2021-11-17
HOCS-4097: make action endpoint explicit for each action

gary-newbolt:


abarker66 (approved):
I can imagine how you feel about having to dumb down your code but I understand why the original comments were made. This will keep the code consistent with what is already in DECS, explicit endpoints and rigid DTOs, etc. It is slightly easier to understand and maintain, if more verbose. It looks good.

PR #478, opened 2021-11-18
Hocs 4176 cpu postgres investigation

nickmcmahon01:
There have been multiple high CPU spikes reported from the 'hocsprodapp' production database. As a result of this there have been a number of high CPU/ slow running queries identified that should be investigated and potentially changed.

Over time the view 'active_case' has become more expensive with added complexity, moving away from the original intent.

Historically this view was used to ensure that child Stages, Topics, Case Notes and Correspondents weren't returned if the parent Case was soft deleted without having to cascade the soft delete to the child entities.

This work moves the existing queries away from using 'active_case' where possible and back to the 'original' intent as the intent of the view.

For this ticket the reduction in query time is almost a side effect, the goal of this work is to reduce the complexity of the queries to reduce the CPU load on the database.


LiamMacP:
You've added queries to somu, don't know if it's quick and dirty to add a deleted column and default to false? I don't mind if you'd prefer it to be a separate change...

LiamMacP (approved):
Please create a ticket for the `somu_item.deleted` column that needs adding. Also will need the Flyway script number changing, latest is V29 I believe.

nickmcmahon01:
> Please create a ticket for the `somu_item.deleted` column that needs adding. Also will need the Flyway script number changing, latest is V29 I believe.

https://collaboration.homeoffice.gov.uk/jira/browse/HOCS-4237

PR #480, opened 2021-11-19
HOCS-4182 Expensive calls to GetStageUser GetStageTeam and GetStageType

nickmcmahon01:
This commit replaces the use of active_case to provide the original intent of the query as active_case has changed.

This change reduces the execution time from 500ms to 0.2ms

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/BaseStage.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/BaseStage.java
@@ -1,0 +1,78 @@
+package uk.gov.digital.ho.hocs.casework.domain.model;
+
+import lombok.Getter;
+import lombok.Setter;
+
+import javax.persistence.Column;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.MappedSuperclass;
+import java.io.Serializable;
+import java.time.LocalDate;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/BaseStage.java
```diff
@@ -0,0 +1,78 @@
...
+    protected UUID userUUID;
+
+    @Getter
+    @Setter
+    @Column(name = "deadline_warning")
+    protected LocalDate deadlineWarning;
+
+    @Getter
+    @Setter
+    @Column(name = "somu", insertable = false, updatable = false)
```
LiamMacP, line 67: I believe this somu is a remnant of an old change. Worth taking it out now we do the processing in the service?

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/Stage.java
```diff
@@ -1,25 +1,28 @@
...
 
 import static uk.gov.digital.ho.hocs.casework.application.LogEvent.STAGE_CREATE_FAILURE;
 
 @NoArgsConstructor
 @Entity
 @Table(name = "stage")
-public class Stage extends AbstractJsonDataMap implements Serializable {
+public class Stage extends BaseStage {
 
     public interface StageTeamUuid {
```
LiamMacP, line 27: This interface can be removed now?

PR #481, opened 2021-11-19
HOCS-4182 Expensive calls to GetStageUser GetStageTeam and GetStageType

nickmcmahon01:
This commit replaces the use of active_case to provide the original intent of the query as active_case has changed.

This change reduces the execution time from 500ms to 0.2ms


nickmcmahon01:
I renamed the branch  and it killed the PR, see comments here;
https://github.com/UKHomeOffice/hocs-casework/pull/480

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
```diff
@@ -355,15 +355,22 @@ private Stage getStage(UUID caseUUID, UUID stageUUID) {
         }
     }
 
+    private BasicStage getBasicStage(UUID caseUUID, UUID stageUUID) {
```
fewtrees-ho, line 358: Name needs to reflect its purpose. Perhaps renaming the other version is more appropriate i.e. `getStageTypeAndCaseFromStageData()` ?

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/BasicStage.java
```diff
@@ -0,0 +1,32 @@
...
+import javax.persistence.Table;
+import java.time.LocalDateTime;
+import java.util.UUID;
+
+import static uk.gov.digital.ho.hocs.casework.application.LogEvent.STAGE_CREATE_FAILURE;
+
+@NoArgsConstructor
+@Entity
+@Table(name = "stage")
+public class BasicStage extends BaseStage {
```
fewtrees-ho, line 16: `BasicStage` vs `BaseStage` is confusing. Why not just `Stage`?

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/Stage.java
```diff
@@ -195,4 +158,12 @@ public boolean isActive() {
         return this.teamUUID != null;
     }
 
+    public void update(Map<String, String> newData, ObjectMapper objectMapper) {
```
fewtrees-ho, line 161: These methods are already part of `AbstractJsonDataMap` so why here?

fewtrees-ho (changes requested):

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/BasicStage.java
nickmcmahon01: I was going for the minimum diff for the PR, happy to rename the current 'Stage' to 'StageData' to reflect that it returns the data blob and 'BasicStage' to 'Stage' ? That would be more correct?

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
nickmcmahon01: Good suggestion

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/Stage.java
nickmcmahon01: We can't extend 2 classes and we need to extend BaseStage as a @MappedSuperclass. I was about to suggest we move AbstractJsonDataMap to a static 'helper' class but I've just noticed it already is 'JsonMapUtils', 'AbstractJsonDataMap' is just a 1 line method to call the helper class so it doesn't need to exist in my opinion - we never pass Stage and CaseData into the same method that uses that base class for example. I'll remove it from both and call the helper class directly...quite why we need a helper class to use a map is for another day 😄

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/BasicStage.java
fewtrees-ho: Or `StageComposite` would suggest there's more in it
`StageWithCase` - is explicit and is without confusion..

`BasicStage` to `Stage` is correct.

fewtrees-ho (approved):
Thanks

PR #487, opened 2021-11-23
HOCS-4220: Add Swagger OpenApi3

LiamMacP:
Recently the ability to spin up Swagger from the
local service port was removed through a mis-configured dependency.
This change incorporates the open source community
led Springdoc dependency that support OpenAPI 3.0.
By default, this is turned off for production but can be enabled
through the 'local' spring profile.


abarker66 (approved):
Looks good, tested and working.

PR #502, opened 2021-12-03
HOCS-4269 - UNCAUGHT_EXCEPTION: Exception: null, Stack: java.lang.Nul…

agraham19:
…lPointerException

For some reason all of the lines after the first are getting cut off the stacktraces
for uncaught exceptions. This problem does not occur in info-service, so this PR
replaces some of the logging lines with a copy and paste from the correctly function
service.


fewtrees-ho (changes requested):
Does this change fix the problem though?

agraham19:
> Does this change fix the problem though?

It attempts to address the logging bug so we can debug the null pointer error.

LiamMacP (approved):
For the issue in question, the correlation ID isn't showing any surrounding logs so we are stabbing in the dark. This should atleast give us the underlying exception. We can then look into potentially why we haven't got surrounding logs...

fewtrees-ho (approved):

PR #510, opened 2021-12-06
Hocs-4235 -'createStage' is using a slow query

nickmcmahon01:
This commit
1) Gets 'createStage' to use the simpler (quicker) repository method without the topics/correspondents/case data.
2) Puts the database code in a transaction as we're modifying two tables
3) Removes multiple calls to getCaseData when one will do.

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
@@ -159,7 +159,7 @@ public class CaseDataService {
     private CaseData getAnyCaseData(UUID caseUUID) {
         log.debug("Getting any Case: {}", caseUUID);
         CaseData caseData = caseDataRepository.findAnyByUuid(caseUUID);
-        if (caseData != null) {
+        if (caseData == null) {
             log.error("Any Case: {}, not found!", caseUUID, value(EVENT, CASE_NOT_FOUND));
             throw new ApplicationExceptions.EntityNotFoundException(String.format("Any Case: %s, not found!", caseUUID), CASE_NOT_FOUND);
         }
--- a/src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
@@ -301,23 +301,23 @@ public class CaseDataService {
     }
 
     public void updateCaseData(UUID caseUUID, UUID stageUUID, Map<String, String> data) {
-        if (data != null) {
-            updateCaseData(getCaseData(caseUUID), stageUUID, data);
-        } else {
+        if (data == null) {
             log.warn("Data was null for Case: {} Stage: {}", caseUUID, stageUUID, value(EVENT, CASE_NOT_UPDATED_NULL_DATA));
+        } else {
+            updateCaseData(getCaseData(caseUUID), stageUUID, data);
         }
     }
 
     public void updateCaseData(CaseData caseData, UUID stageUUID, Map<String, String> data) {
         log.debug("Updating data for Case: {}", caseData.getUuid());
-        if (data != null) {
+        if (data == null) {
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
```diff
@@ -189,10 +189,14 @@ public String getCaseDataCaseRef(UUID caseUUID) {
     }
 
     public String getCaseDataField(UUID caseUUID, String key) {
-        log.debug("Looking up key {} for Case: {}", key, caseUUID);
-        Map<String, String> dataMap = getCaseData(caseUUID).getDataMap(objectMapper);
+        return getCaseDataField(getCaseData(caseUUID), key);
+    }
+
+    public String getCaseDataField(CaseData caseData, String key) {
```
LiamMacP, line 195: Is this abstraction actually necessary, Line 308 does the same `getCaseData` call. Worth just moving this within the UUID for now. We can always add this back in the future...

src/main/java/uk/gov/digital/ho/hocs/casework/api/dto/CreateStageResponse.java
```diff
@@ -15,7 +15,7 @@
     @JsonProperty("uuid")
     private final UUID uuid;
 
-    public static CreateStageResponse from(StageWithCaseData stage) {
+    public static CreateStageResponse from(BaseStage stage) {
```
LiamMacP, line 18: This response literally only needs a UUID, I see why you haven't done it to minimize the diff, but feels silly chucking the whole object around...

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
```diff
@@ -302,16 +306,21 @@ private CaseData createCaseBaseData(String caseType, Map<String, String> data, L
     }
 
     public void updateCaseData(UUID caseUUID, UUID stageUUID, Map<String, String> data) {
-        log.debug("Updating data for Case: {}", caseUUID);
+        if (data != null) {
```
LiamMacP, line 309: If data is null, we don't log, ideal?

LiamMacP (approved):

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
```diff
@@ -302,16 +306,21 @@ private CaseData createCaseBaseData(String caseType, Map<String, String> data, L
...
+            updateCaseData(getCaseData(caseUUID), stageUUID, data);
+        }
+    }
+
+    public void updateCaseData(CaseData caseData, UUID stageUUID, Map<String, String> data) {
+        log.debug("Updating data for Case: {}", caseData.getUuid());
         if (data != null) {
```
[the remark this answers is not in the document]
nickmcmahon01: I did that and put it back. I'll do it then :)

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
```diff
@@ -126,50 +126,45 @@ public CaseData getCase(UUID caseUUID) {
...
             throw new ApplicationExceptions.EntityNotFoundException(String.format("Case: %s, not found!", reference), CASE_NOT_FOUND);
         }
+        log.info("Got Case by reference: {}", caseData.getUuid(), value(EVENT, CASE_RETRIEVED));
+        return caseData;
     }
 
     private CaseData getAnyCaseData(UUID caseUUID) {
         log.debug("Getting any Case: {}", caseUUID);
         CaseData caseData = caseDataRepository.findAnyByUuid(caseUUID);
         if (caseData != null) {
```
LiamMacP, line 162: This should be `==`?

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
LiamMacP: Could we do the same here for the `if == null`?

PR #511, opened 2021-12-06
Hocs-4235 -'getTeams' should use the existing getBasicStage method

nickmcmahon01:
This commit moves 'getTeams' to use the existing getBasicStage method, this isn't an important change on its own but it is the first part of being able to return a 'Stage' object in callers to the Casework service instead of calling it n times in one method for specific fields as we do now.

It also puts back an exception where there stageUUID is not found in the database - this is different to 'inactive' stages which don't have teams assigned to them

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
@@ -92,6 +92,7 @@ public class StageService {
     }
 
     public UUID getStageUser(UUID caseUUID, UUID stageUUID) {
+        // Should call Active stage because to have a user assigned there should also be a team assigned.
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
```diff
@@ -92,6 +92,7 @@ public StageService(StageRepository stageRepository, UserPermissionsService user
     }
 
     public UUID getStageUser(UUID caseUUID, UUID stageUUID) {
+        // Should call Active stage because to have a user assigned there should also be a team assigned.
```
LiamMacP, line 95: I'd prefer more of a javadoc type function comment but minor...

LiamMacP (approved):

PR #512, opened 2021-12-06
Hocs-4235 -'getRef' is always called when we have access to the ref already

nickmcmahon01:
This commit removes 'getCaseRef' as every time it is called we already have access to the case reference and we don't need to fetch it again from the case_data table.


nickmcmahon01:
Its probably worth merging the other ones before reviewing this as some of the previous code needs to be in place for it make sense.

PR #517, opened 2021-12-15
HOCS-4198 Extra blank case note when adding/updating appeals and 'System Event' log in timeline when an extension is applied

agraham19:
* Stop saving case notes for actions
* Rely on the audit timeline for action notes


abarker66 (approved):
LGTM, system tests pass.

PR #529, opened 2021-12-17
HOCS-4373: rework user permissions

LiamMacP:
Presently, a lot of our checks revolve around getting a users teams and
then calculating the case types that user is a case admin for. While
this works, if we ever need them teams we have to search for all the
teams for them case types separately. This change offers an 'expanded'
view of a users teams whereby all the team uuids that a user is a member
of, or a case admin of are added together. This will support future
permission development for ensuring users have the correct access to
workstacks and teams.

A side effect of this change is we gain a performance bonus by removing
the case type check this simplifies the queries that used it. From
`12329.683ms` to `731.688ms` on a production dataset.


src/main/java/uk/gov/digital/ho/hocs/casework/security/UserPermissionsService.java
```diff
@@ -32,31 +33,54 @@ public UUID getUserId() {
...
+    public Set<UUID> getExpandedUserTeams() {
+        var userTeamUuids = getUserTeams();
 
+        if (userTeamUuids.isEmpty()) {
+            return userTeamUuids;
+        }
 
+        var allTeams = infoClient.getTeams();
+
+        var userCaseAdminCaseTypes = getUserCaseAdminCaseTypes(allTeams, getUserTeams());
```
nickmcmahon01, line 65: not blocking, but can we call getUserTeams once in this method?

src/main/java/uk/gov/digital/ho/hocs/casework/api/DashboardService.java
```diff
@@ -37,20 +37,13 @@ public DashboardService(DashboardSummaryFactory dashboardSummaryFactory,
     Map<UUID, Map<String, Integer>> getDashboard() {
         log.debug("Getting dashboard summary for user");
 
-        List<Summary> casesSummaries = new ArrayList<>();
-
-        Set<UUID> allUserTeams = userPermissionsService.getUserTeams();
-        if(!allUserTeams.isEmpty()) {
-            casesSummaries.addAll(summaryRepository.findTeamsAndCaseCountByTeamUuid(allUserTeams));
-        } else {
+        Set<UUID> allUserTeams = userPermissionsService.getExpandedUserTeams();
```
nickmcmahon01, line 40: This is much simpler!

src/main/java/uk/gov/digital/ho/hocs/casework/security/UserPermissionsService.java
```diff
@@ -1,25 +1,26 @@
...
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 @Service
-@Slf4j
 public class UserPermissionsService {
 
-    private RequestData requestData;
-    private InfoClient infoClient;
+    private final RequestData requestData;
```
nickmcmahon01, line 21: 👮 I'm glad the final police have swung by. this is the most important change in this PR IMO

src/main/java/uk/gov/digital/ho/hocs/casework/security/UserPermissionsService.java
LiamMacP: I put it above for reuse, not sure why I didn’t change this. Will sort it later!

nickmcmahon01 (approved):
Approved, but did we not agree to add the permissions check to the method(s) that are missing them at the moment?

src/main/java/uk/gov/digital/ho/hocs/casework/security/UserPermissionsService.java
LiamMacP: Changed...

src/main/java/uk/gov/digital/ho/hocs/casework/security/UserPermissionsService.java
LiamMacP: Nee noor nee noor 👮🏽

PR #530, opened 2021-12-18
♻️ HOCS-4383: Create a CaseDataTypeService

nickmcmahon01:
As part of decoupling our microservices we need to start to treat data that is currently mastered in Info service as part of the casework domain.
This PR is a refactor to create a CaseDataTypeService which initially retrieves data from Info service and later will be changed to use a repository class without having to change the rest of the code. This code has moved from CaseDataService and StageService.


src/main/java/uk/gov/digital/ho/hocs/casework/client/infoclient/InfoClient.java
```diff
@@ -30,24 +30,9 @@ public InfoClient(RestHelper restHelper,
         this.serviceBaseURL = infoService;
     }
```
nickmcmahon01, line 32: We can reduce this to one call that fetches all caseDataTypes - we can present it differently in the service layer. it also allows us to deprecate or remove these calls from info-service.

src/main/java/uk/gov/digital/ho/hocs/casework/migration/MigrationCaseDataService.java
```diff
@@ -56,7 +56,7 @@ CaseData createCase(MigrationCreateCaseRequest request) {
         String totalsListName = request.getTotalsListName();
 
         log.debug("Creating Case of type: {}", caseType);
-        CaseDataType caseDataType = infoClient.getCaseType(caseType);
+        CaseDataType caseDataType = null;//infoClient.getCaseType(caseType);
```
nickmcmahon01, line 59: 👎

src/test/java/uk/gov/digital/ho/hocs/casework/api/CaseDataCreateCaseIntegrationTest.java
```diff
@@ -70,7 +70,7 @@ public void shouldCreateACaseWithPermissionLevelOwner() throws JsonProcessingExc
         long numberOfCasesBefore = caseDataRepository.count();
 
         setupMockTeams("TEST", 5);
```
nickmcmahon01, line 72: Boyscout changes, we were passing in these params and not using it them, it looks like they've moved to setupMockTeams above and never been removed from this method params

src/test/java/uk/gov/digital/ho/hocs/casework/api/CaseDataCreateCaseIntegrationTest.java
```diff
@@ -296,9 +296,9 @@ private void setupMockTeams(String caseType, int permission) throws JsonProcessi
         teamDtos.add(teamDto);
 
         mockInfoService
-                .expect(requestTo("http://localhost:8085/caseType/type/TEST"))
+                .expect(requestTo("http://localhost:8085/caseType"))
```
nickmcmahon01, line 299: Most of the real test changes are repointing wiremock to use the one remaining endpoint.

src/test/java/uk/gov/digital/ho/hocs/casework/api/CaseDataServiceTest.java
```diff
@@ -152,7 +138,7 @@ public void shouldCreateCase() throws ApplicationExceptions.EntityCreationExcept
 
         when(caseDataRepository.getNextSeriesId()).thenReturn(caseID);
         when(infoClient.getCaseDeadline(caseType.getDisplayCode(), deadlineDate, 0)).thenReturn(caseDeadline);
-        when(infoClient.getCaseType(caseType.getDisplayCode())).thenReturn(caseType);
```
nickmcmahon01, line 155: Where we were calling infoClient directly we have repointed the code to call the new CaseDataTypeService.

src/test/java/uk/gov/digital/ho/hocs/casework/api/CaseDataServiceTest.java
```diff
@@ -899,42 +885,6 @@ public void shouldNotDeleteCaseMissingCaseUUID() {
         verifyNoMoreInteractions(caseDataRepository);
     }
 
-    @Test
```
nickmcmahon01, line 902: These have been removed from CaseDataServiceTest and moved to CaseDataTypeServiceTest

src/test/java/uk/gov/digital/ho/hocs/casework/security/AuthorisationAspectTest.java
```diff
@@ -58,167 +58,149 @@ public void setup() {
...
         request.setMethod("GET");
         RequestContextHolder.setRequestAttributes(new ServletRequestAttributes(request));
-        aspect = new AuthorisationAspect(caseService, userService);
+        aspect = new AuthorisationAspect(caseService, caseDataTypeService, userService);
     }
 
     @Test
     public void shouldCaseServiceLookupWhenExistingCase() throws Throwable {
 
-        String type = "MIN";
```
nickmcmahon01, line 67: The gitblame is me, but I don't like using real data in tests where possible, these tests are not DCU specific behaviour

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataTypeService.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private final InfoClient infoClient;
+
+    @Autowired
+    public CaseDataTypeService(InfoClient infoClient) {
+        this.infoClient = infoClient;
+    }
+
+    public Stream<CaseDataType> getAllCaseDataTypes(Predicate<CaseDataType> predicate) {
+        return infoClient.getAllCaseDataTypes().stream().filter(predicate);
```
LiamMacP, line 29: Part of me sort of feels like the Predicate stuff shouldn't be part of this method, `getAllCaseDataTypes` aludes to returning them all...

LiamMacP: I do however really like us using a properly cached method and pulling the value from that...

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataTypeService.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public CaseDataType getCaseDataType(String caseType) {
+        var type = getNonNullValue(caseType);
+        return getAllCaseDataTypes(it -> it.getDisplayCode().equals(type)).findFirst().orElseThrow(() -> getNotFoundException(type));
+    }
+
+    public CaseDataType getCaseDataType(UUID caseUUID){
+        var shortCode = getNonNullValue(caseUUID).toString().substring(34);
```
LiamMacP, line 38: Hmm, can we not move away from the weird manipulation of UUIDs? I know why they are there, it just feels wrong though...

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataTypeService.java
```diff
@@ -0,0 +1,50 @@
...
+        return getAllCaseDataTypes(it -> it.getShortCode().equals(shortCode)).findFirst().orElseThrow(() -> getNotFoundException(shortCode));
+    }
+
+    private static <T> T getNonNullValue(T value) {
+        return Optional.ofNullable(value).orElseThrow(() -> getNotFoundException("null"));
```
LiamMacP, line 43: Wouldn't this always just lead to `CaseDataType: null, not found!`, not sure thats much use...

src/main/java/uk/gov/digital/ho/hocs/casework/api/TopicResource.java
```diff
@@ -56,15 +56,9 @@ ResponseEntity deleteTopic(@PathVariable UUID caseUUID, @PathVariable UUID stage
         return ResponseEntity.ok().build();
     }
 
-    @PostMapping(value = "/topic/{topicUUID}/clearCachedStandardLine")
-    ResponseEntity clearCachedStandardLineForTopic(@PathVariable UUID topicUUID) {
```
LiamMacP, line 60: Glad to see this go if unused...

src/test/java/uk/gov/digital/ho/hocs/casework/api/CaseDataCreateCaseIntegrationTest.java
```diff
@@ -48,7 +48,7 @@
 @ActiveProfiles({ "local", "integration" })
 public class CaseDataCreateCaseIntegrationTest {
     private MockRestServiceServer mockInfoService;
-    private TestRestTemplate testRestTemplate = new TestRestTemplate();
+    private final TestRestTemplate testRestTemplate = new TestRestTemplate();
```
LiamMacP, line 51: 👮🏽‍♂️

src/test/java/uk/gov/digital/ho/hocs/casework/api/StageServiceTest.java
```diff
@@ -64,7 +66,7 @@
     private final String allocationType = "anyAllocate";
     private final UUID transitionNoteUUID = UUID.randomUUID();
     private final CaseDataType caseDataType = new CaseDataType("MIN", "1a", "MIN", null);
-    private final List<CaseDataType> caseDataTypes = List.of(
+    private final Stream<CaseDataType> caseDataTypes = Stream.of(
             CaseDataTypeFactory.from("NXT", "a5", "MIN"), // NXT can be reached through MIN
```
LiamMacP, line 70: Not really related but can we change `MIN` to something not tenant specifc?

src/test/java/uk/gov/digital/ho/hocs/casework/security/SecurityIntegrationTest.java
```diff
@@ -53,6 +54,9 @@
     @Qualifier("CaseDataService")
     CaseDataService caseDataService;
 
+    @MockBean()
```
LiamMacP, line 57:
```suggestion
    @MockBean
```

LiamMacP (dismissed):
Nothing that cannot be re-worked going forward, good improvement overall...

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataTypeService.java
nickmcmahon01: I agree, but it needs fixing upstream of here, for the security code we can stop looking up the caseType completely and fetch the actual case from the database, which is quick enough, but then semantically there is a difference 'if it exists this UUID would be a MIN' to 'this case isn't in the database' so we would need to ensure the behaviour stays the same.

PR #533, opened 2021-12-21
HOCS-4008: team stage endpoint checks user teams

LiamMacP:
This change amends the team endpoint to check that a user is part of the
required team to return the data. If the user is notpart of a required
team then the service throws a forbidden exception and returns a 403
status code.


src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
```diff
@@ -261,6 +265,14 @@ void updateStageUser(UUID caseUUID, UUID stageUUID, UUID newUserUUID) {
 
     Set<StageWithCaseData> getActiveStagesByTeamUUID(UUID teamUUID) {
         log.debug("Getting Active Stages for Team: {}", teamUUID);
+
+        Set<UUID> usersTeam = userPermissionsService.getExpandedUserTeams();
+
+        if (!usersTeam.contains(teamUUID)) {
+            log.warn("User {} attempted to view team {}", userPermissionsService.getUserId(), teamUUID, value(EVENT, SECURITY_FORBIDDEN));
```
nickmcmahon01, line 272: 👍

nickmcmahon01 (approved):

nickmcmahon01:
This looks like it could be a good annotation, but we'd also want to use the teamuuids directly so it seems a shame to have to do that work twice.

LiamMacP:
> This looks like it could be a good annotation, but we'd also want to use the teamuuids directly so it seems a shame to have to do that work twice.

In this case we'd only need the annotation, but for some of the others we'd have to do it twice. Maybe that's the price we will have to pay for simplicity...

PR #539, opened 2021-12-22
HOCS-3953: remove camel from service

LiamMacP:
This change works to refactor all usages of camel to AWS provided SDKs to remove Camel from the service. 

To support this the following changes have been incorporated into this change: 

1. Pipeline and local dev changes - adds localstack instantiation to Drone and initialises required resources
2. Removes unnecessary and no longer required camel dependencies
3. Reworks the kubernetes deployment config to support now changed application.properties
4. Removed Camel processor from the AuditClient and NotifyClient, replacing the usage with the AWS SDKs
5. Added proper integration tests to ensure the messages we send are what we expect

A number of boyscout changes have also been incorporated: 

1. Simplify AWS SQS and SNS bean creation
2. Removed unrequired property files
3. Fixed incorrect test cases


build.gradle
```diff
@@ -53,8 +51,6 @@ dependencies {
     implementation('org.springdoc:springdoc-openapi-ui:1.6.2')
 
     implementation('com.github.ben-manes.caffeine:caffeine:3.0.5')
-    implementation("org.apache.camel:camel-spring-boot-starter:${camelVersion}")
-    implementation("org.apache.camel:camel-aws:${camelVersion}")
     implementation('com.amazonaws:aws-java-sdk:1.12.131')
```
nickmcmahon01, line 54: Can this go? its hundreds of Mb
See https://github.com/UKHomeOffice/hocs-casework/pull/540
Screenshot 2021-12-22 at 22 31 38

docker-compose.yml
```diff
@@ -16,45 +16,17 @@ services:
       - /var/lib/postgresql
 
   localstack:
-    image: localstack/localstack:latest
+    image: localstack/localstack:0.11.1
```
nickmcmahon01, line 19: This needs doing  e v e r y w h e r e

kube/kd/deployment.yaml
```diff
@@ -266,18 +266,23 @@ spec:
                 secretKeyRef:
                   name: hocs-casework
                   key: rds_password
-            - name: AUDIT_TOPIC_NAME
+            - name: AWS_SNS_AUDIT_SEARCH_TOPIC_NAME
```
nickmcmahon01, line 269: Oh look, we're deploying otherwise breaking changes to the deployment along with the code that needs it :)

src/main/java/uk/gov/digital/ho/hocs/casework/client/notifyclient/NotifyClient.java
```diff
@@ -1,94 +1,90 @@
...
+                new TeamAssignChangeCommand(caseUUID, stageUUID, caseReference, teamUUID, emailType),
+                LogEvent.TEAM_EMAIL_SENT);
     }
 
-    public void sendUserEmail(UUID caseUUID, UUID stageUUID, UUID currentUserUUID, UUID newUserUUID, String caseReference)  {
-        sendUserEmailCommand(new UserAssignChangeCommand(caseUUID, stageUUID, caseReference, currentUserUUID, newUserUUID));
+    public void sendUserEmail(UUID caseUUID, UUID stageUUID, UUID currentUserUUID, UUID newUserUUID, String caseReference) {
+        sendMessage(
+                new UserAssignChangeCommand(caseUUID, stageUUID, caseReference, currentUserUUID, newUserUUID),
+                LogEvent.USER_EMAIL_SENT);
```
nickmcmahon01, line 58: Good spot

src/main/java/uk/gov/digital/ho/hocs/casework/util/SnsStringMessageAttributeValue.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.digital.ho.hocs.casework.util;
```
nickmcmahon01, line 1: How do we do this elsewhere? is this really a problem?

src/main/java/uk/gov/digital/ho/hocs/casework/util/SqsStringMessageAttributeValue.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.digital.ho.hocs.casework.util;
+
```
nickmcmahon01, line 2: We don't need two of these there is nothing 'sqs' about this and 'sns' about the other

src/test/resources/application-integration.properties
```diff
@@ -0,0 +1 @@
+spring.cache.type=NONE
```
nickmcmahon01, line 1: can this just be in an application properties in the test resources folder?

src/main/java/uk/gov/digital/ho/hocs/casework/util/SqsStringMessageAttributeValue.java
LiamMacP: The MessageAttributeValue class is actually an import from the specific SQS and SNS library if you check the import. Pain in the backside but I'd rather we have both incase they breaking change one in the future.

src/test/resources/application-integration.properties
LiamMacP: You then have to copy the whole of the application.properties file again as it overwrites. I've chosen this route so we don't have to copy everything. Happy for you to disagree though...

src/main/java/uk/gov/digital/ho/hocs/casework/util/SnsStringMessageAttributeValue.java
LiamMacP: 'hocs-case-creator' has this class, there was a comment on the remove camel PR to make it a class to simplify the instantiating.

build.gradle
LiamMacP: Good catch, yes will sort tomorrow...

src/main/java/uk/gov/digital/ho/hocs/casework/util/SqsStringMessageAttributeValue.java
nickmcmahon01: 😢 👍

src/test/resources/application-integration.properties
nickmcmahon01: 😢 👍

nickmcmahon01 (approved):

PR #540, opened 2021-12-22
HOCS-3953: replace AWS libs for just messaging ones

nickmcmahon01:


LiamMacP (approved):
Change looks good to me!

PR #549, opened 2022-01-05
HOCS-4418 Fix tag decoration on stage service

robertdeniszczyc2:
Correctly set returned tags onto stages to return to the frontend

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/StageWithCaseData.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/StageWithCaseData.java
@@ -133,10 +133,13 @@ public class StageWithCaseData extends BaseStage {
     }
 
     public void putData(String key, String value) {
+        if (this.data == null) {
+            this.data = new HashMap<>();
+        }
+
         var newData = new HashMap<>(this.data);
         newData.put(key, value);
         this.data = newData;
-
     }
 
     public void putAllData(Map<String,String> data) {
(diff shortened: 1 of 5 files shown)


LiamMacP:
Is there any tests that can be added to facilitate future changes?

robertdeniszczyc2:
> Is there any tests that can be added to facilitate future changes?

Updated with a test, also had to make a change to src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/StageWithCaseData.java to avoid a null pointer exception when putting data potential scenario you might also want to have a look at please

Thankyou!

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/StageWithCaseData.java
```diff
@@ -133,10 +133,13 @@ public StageWithCaseData(UUID caseUUID, String stageType, UUID teamUUID, UUID us
     }
 
     public void putData(String key, String value) {
+        if (this.data == null) {
+            this.data = new HashMap<>();
+        }
+
         var newData = new HashMap<>(this.data);
```
LiamMacP, line 140: Is this a little redundant? could we not just do `data.put` here and remove this?

robertdeniszczyc2: You wouldn't be able to do data.put if this.data was null, so it would need to be instantiated first

LiamMacP: Sorry I didn't explain there well. We keep the null check and then just add into the map instead of having a second map.

LiamMacP (approved):
Changes looks good to me!

PR #570, opened 2022-02-03
HOCS-4421: deployment uses point in time config

LiamMacP:
Releases that are currently handled through the pipeline use the
kubernetes configuration at the head of main and not the point in time
for that 'version'. This change includes adding a 'fetch and checkout'
step that checks out the Git commit SHA for dev deployments and the
SemVar tag for other releases. This enables us to use the config from
that point in time.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -58,7 +58,6 @@ steps:
       registry: quay.io
       repo: quay.io/ukhomeofficedigital/hocs-casework
       tags:
-        - build_${DRONE_BUILD_NUMBER}
         - ${DRONE_COMMIT_SHA}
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -117,6 +115,11 @@ environment:
   DOCKER_HOST: tcp://docker:2375
 
 steps:
+  - name: fetch and checkout
+    image: alpine/git
+    commands:
+      - git fetch --tags
+      - git checkout $${VERSION}
```
zuzak, line 122: Will this break if we use `build_123`?

LiamMacP: Yes this would break - we are no longer using build_ and now using the commit SHAs. We don't push the build_ tag anymore.

PR #578, opened 2022-02-18
HOCS-4185 BF2 case type

robertdeniszczyc2:


gary-newbolt (approved):
Happy to approve as it's fairly simple method, but did you want to add a test on the new Copy class?

gary-newbolt (approved):
lgtm

PR #589, opened 2022-02-28
HOCS-4525 Add PreviousCaseReference into BF2 case data

robertdeniszczyc2:


LiamMacP (approved):
Change looks fine - approved

PR #598, opened 2022-03-08
HOCS-4724: send final stage completed audit

LiamMacP:
When a case is at the end of a workflow, when completing a case we don't
send a STAGE_COMPLETED audit line for that stage. This change triggers
this audit event when a case is completed with an active stage.


gary-newbolt (approved):
lgtm

PR #601, opened 2022-03-11
HOCS-4682: pass through document values

LiamMacP:
Document contain two booleans in the response payload that indicate if a
file or PDF file exists. This changes passes this through to the REST
response for these calls.


nickmcmahon01:
Label police, does this change big enough to move to hocs-casework 2.0?

robertdeniszczyc2 (approved):

PR #605, opened 2022-03-15
HOCS-4716 BF2 Field auto population fix

robertdeniszczyc2:
Removes case summary, previous complaint reference, third party reference
from field auto population on BF2 case creation

### DIFF
--- a/src/test/java/uk/gov/digital/ho/hocs/casework/api/factory/strategies/CopyBfToBf2Test.java
+++ b/src/test/java/uk/gov/digital/ho/hocs/casework/api/factory/strategies/CopyBfToBf2Test.java
@@ -28,12 +28,12 @@ public class CopyBfToBf2Test {
 
     static {
         FROM_CLOB = Map.of(
-                "CaseSummary", "TestValue");
+                "ComplaintReason1", "TestValue");
     }
 
     static {
         TO_CLOB = Map.of(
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/digital/ho/hocs/casework/api/factory/strategies/CopyBfToBf2Test.java
```diff
@@ -28,12 +28,12 @@
 
     static {
         FROM_CLOB = Map.of(
-                "CaseSummary", "TestValue");
+                "ComplaintReason1", "TestValue");
```
LiamMacP, line 31: `ComplaintReason1` related to this change?

Also sort of feels like you should have the caseSummary etc in this.

LiamMacP (commented):
Just a couple of comments

src/test/java/uk/gov/digital/ho/hocs/casework/api/factory/strategies/CopyBfToBf2Test.java
robertdeniszczyc2: Will also resolve what should be in FROM_CLOB

robertdeniszczyc2:
> Just a couple of comments 🙂

Updated for comments, cheers!

LiamMacP (approved):
Changed handled and looks good - approved!

PR #614, opened 2022-03-24
Hocs-4799: Update the base image to fix the drone build

nickmcmahon01:


LiamMacP (approved):
Changes look good - approved

PR #624, opened 2022-04-06
HOCS-3803: add suspend feature for smc cases

gary-newbolt:


src/main/java/uk/gov/digital/ho/hocs/casework/api/ActionDataSuspendService.java
```diff
@@ -0,0 +1,162 @@
...
+                    )).collect(Collectors.toList());
+    }
+
+    public String suspend(UUID caseUUID, UUID existingStageUuid, ActionDataSuspendDto suspendDto) {
+
+        log.debug("Request to suspend case {} received", caseUUID);
+
+        CaseData caseData = caseDataRepository.findActiveByUuid(caseUUID);
+        if (caseData == null) {
+            throw new ApplicationExceptions.EntityNotFoundException(String.format("Case with id: %s does not exist.", caseUUID), CASE_NOT_FOUND);
```
robertdeniszczyc2, line 79:
```suggestion
            throw new ApplicationExceptions.EntityNotFoundException(String.format("Case with uuid: %s does not exist.", caseUUID), CASE_NOT_FOUND);
```

src/main/java/uk/gov/digital/ho/hocs/casework/api/utils/ActionDataHelpers.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.time.LocalDate;
+import java.util.Map;
+
+@Slf4j
+public class ActionDataHelpers {
+    private ActionDataHelpers() {}
+
+    // COPIED FROM CaseDataService to avoid cyclic dependency.
+    public static void updateStageDeadlines(CaseData caseData) {
```
robertdeniszczyc2, line 15: This is probably better to implement as non-void and include tests written against it

src/main/java/uk/gov/digital/ho/hocs/casework/api/ActionDataSuspendService.java
```diff
@@ -0,0 +1,162 @@
...
+            String msg = String.format("No suspension action exists for id %s in case %s", existingSuspensionUUID, caseUUID);
+            log.error(msg);
+            throw new ApplicationExceptions.EntityNotFoundException(msg, LogEvent.ACTION_DATA_UPDATE_FAILURE);
+        });
+
+        log.info("Suspension {} for caseId {} has been removed.", existingSuspensionUUID, caseUUID);
+        return caseData.getReference();
+    }
+
+    private boolean hasMaxActiveRequests(CaseTypeActionDto caseTypeActionDto, UUID caseUUID) {
```
robertdeniszczyc2, line 158: Not sure this is required

src/main/java/uk/gov/digital/ho/hocs/casework/api/ActionDataSuspendService.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        // Save the suspension event.
+        ActionDataSuspension suspensionEntity = new ActionDataSuspension(
+                caseTypeActionDto.getUuid(),
+                caseTypeActionDto.getActionSubtype(),
+                caseTypeActionDto.getActionLabel(),
+                caseData.getType(),
+                caseUUID,
+                suspendDto.getDateSuspensionApplied(),
```
robertdeniszczyc2, line 100: This field is defined as a timestamp but we only pass a date into it, which means all the records have a time of 00:00:00 set against them which is inaccurate. Would be good to revisit this field to maintain accuracy.

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseActionsResource.java
```diff
@@ -87,4 +96,24 @@ public ResponseEntity<GetCaseReferenceResponse> updateExternalInterest(@PathVari
...
+    public ResponseEntity<GetCaseReferenceResponse> applyCaseSuspension(@PathVariable UUID caseUUID,
+                                                                        @PathVariable UUID stageUUID,
+                                                                        @RequestBody ActionDataSuspendDto suspendDto) {
+
+        String caseRef = suspensionService.suspend(caseUUID, stageUUID, suspendDto);
+        return ResponseEntity.ok(GetCaseReferenceResponse.from(caseUUID, caseRef));
+    }
+
+
+    @DeleteMapping(path = "/case/{caseUUID}/stage/{stageUUID}/actions/suspension/{suspensionUUID}")
```
robertdeniszczyc2, line 110: Not sure delete mapping is the right mapping here, when we unsuspend a case we're not deleting any records as far as I can tell. If we are just updating the data_suspension_removed then PUT probably more suited here since we are updating a record

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
```diff
@@ -579,15 +579,18 @@ CaseSummary getCaseSummary(final UUID caseUUID) {
...
+                .withPreviousCaseStageUUID(activeCaseViewData.getPreviousCaseStageUUID());
+
+        if (caseDataMap.containsKey("suspended")) {
+            caseSummary.withSuspended(caseDataMap.get("suspended"));
+        }
 
         log.info("Got Case Summary for Case: {} Ref: {}", caseData.getUuid(), caseData.getReference(), value(EVENT, CASE_SUMMARY_RETRIEVED));
 
-        return caseSummary;
+        return caseSummary.build();
```
robertdeniszczyc2, line 593: As we're now running build() here, the log statement logs we have got the case summary before it's returned from the builder

src/main/java/uk/gov/digital/ho/hocs/casework/api/utils/ActionDataHelpers.java
```diff
@@ -0,0 +1,40 @@
...
+import uk.gov.digital.ho.hocs.casework.domain.model.CaseData;
+
+import java.time.LocalDate;
+import java.util.Map;
+
+@Slf4j
+public class ActionDataHelpers {
+    private ActionDataHelpers() {}
+
+    // COPIED FROM CaseDataService to avoid cyclic dependency.
```
robertdeniszczyc2, line 14: If it's an exact copy can CaseDataService use this class aswell?

src/main/java/uk/gov/digital/ho/hocs/casework/client/auditclient/AuditClient.java
```diff
@@ -617,5 +620,35 @@ public void addAppealOfficerInfo(AuditPayload.AppealItem.AppealItemBuilder build
...
+                suspensionEntity.getActionSubtype(),
+                suspensionEntity.getCaseTypeActionLabel(),
+                suspensionEntity.getCaseDataType(),
+                suspensionEntity.getCaseDataUuid(),
+                suspensionEntity.getDateSuspensionApplied(),
+                suspensionEntity.getDateSuspensionRemoved(),
+                EventType.CASE_UNSUSPENDED
+        );
+        saveActionAudit(payload);
+    }
```
robertdeniszczyc2, line 653: These are identical aside from the EventType, could combine them and have a second parameter to flip between applying suspension or unsuspending

src/main/java/uk/gov/digital/ho/hocs/casework/domain/repository/SuspensionRepository.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.Optional;
+import java.util.UUID;
+
+public interface SuspensionRepository extends CrudRepository<ActionDataSuspension, UUID> {
+
+    Optional<List<ActionDataSuspension>> findAllByCaseDataUuid(UUID caseUUID);
+
+    Optional<ActionDataSuspension> findByUuidAndCaseDataUuid(UUID uuid, UUID caseUUID);
+
+    Optional<List<ActionDataSuspension>> findAllByCaseDataUuidAndCaseTypeActionUuidAndDateSuspensionRemovedIsNull(UUID caseUUID, UUID caseActionTypeUUID);
```
robertdeniszczyc2, line 16: Only usage of this is in a method I'm not sure is needed so same applies to this

src/main/resources/db/migration/postgresql/V1_39__3803_ADD_ACTION_SUSPENSION_TABLE.sql
```diff
@@ -0,0 +1,16 @@
+SET search_path TO casework;
+
+CREATE TABLE action_data_suspensions (
```
robertdeniszczyc2, line 3:
```suggestion
CREATE TABLE IF NOT EXISTS action_data_suspensions (
```

src/main/resources/db/migration/postgresql/V1_39__3803_ADD_ACTION_SUSPENSION_TABLE.sql
```diff
@@ -0,0 +1,16 @@
...
+    uuid                        UUID PRIMARY KEY,
+    action_uuid                 UUID NOT NULL, -- knitted to info.schema case_type_action table entry uuid.
```
robertdeniszczyc2, line 5: Is there a potential for separate piece here to move case_type_action out of info maybe?

src/test/java/uk/gov/digital/ho/hocs/casework/api/ActionDataSuspendServiceTest.java
```diff
@@ -0,0 +1,347 @@
...
+        ActiveStage stage1 = new ActiveStage();
+        ActiveStage stage2 = new ActiveStage();
+        stage1.setDeadline(ORIGINAL_CASE_DEADLINE.minusDays(3));
+        stage1.setDeadlineWarning(ORIGINAL_CASE_DEADLINE.minusDays(5));
+        stage2.setDeadline(ORIGINAL_CASE_DEADLINE.minusDays(2));
+        stage2.setDeadlineWarning(ORIGINAL_CASE_DEADLINE.minusDays(4));
+
+        Set<ActiveStage> activeStages = Set.of(stage1, stage2);
+
+        EXISTING_CASE =  new CaseData(
```
robertdeniszczyc2, line 166:
```suggestion
        EXISTING_CASE = new CaseData(
```

robertdeniszczyc2 (changes requested):
Added a few comments, any questions give me a shout,

Cheers,
Rob

src/main/java/uk/gov/digital/ho/hocs/casework/api/utils/ActionDataHelpers.java
LiamMacP: Better yet maybe a CaseDeadlineHelper that they both use?

gary-newbolt: Ok... so when have looked at this, the method is somewhat different now from when it was originally taken from the `CaseDataService` (August last year). So I've renamed the method to distinguish and changed the class name to  Liam's suggestion.

gary-newbolt: I've also removed some logic that is not required for either Extensions or Suspensions.

src/main/java/uk/gov/digital/ho/hocs/casework/api/utils/ActionDataHelpers.java
gary-newbolt: I've added a test, however I've left the method as void, it isn't necessary to return an object for the test and I think leaving as void is a clearer indication that we are manipulating an existing object rather than returning the passed in object, that from the outside looks like creating a new CaseData object.

src/main/java/uk/gov/digital/ho/hocs/casework/api/ActionDataSuspendService.java
gary-newbolt: good point, since only the date is required, I've altered the DB field to be type `DATE`.

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
gary-newbolt: Altered, assigned the built CaseSummaryResponse to a var before the log, then returned that var.

src/main/java/uk/gov/digital/ho/hocs/casework/client/auditclient/AuditClient.java
gary-newbolt: I've condensed the building of the payload to a separate method to reduce code duplication, however I've avoided creating a single public audit method, so as to fit the other patterns used in this client.

src/main/resources/db/migration/postgresql/V1_39__3803_ADD_ACTION_SUSPENSION_TABLE.sql
gary-newbolt: I think that possible however is probably dependent on the future architecture of DECS conversation. I'm not sure I would want to do any of it right now.

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseActionsResource.java
gary-newbolt: Development hangover... I've updated to a PUT request

src/main/java/uk/gov/digital/ho/hocs/casework/domain/repository/SuspensionRepository.java
gary-newbolt: I'll need to leave this in regardless as to whether the method that wraps it remains or not as there will need to be a check so as not to apply two suspensions to the deadline either way.

robertdeniszczyc2 (approved):

PR #668, opened 2022-05-23
Start to improve the slowest query in the system

nickmcmahon01:


src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
```diff
@@ -128,7 +128,7 @@ public CaseData getCase(UUID caseUUID) {
         return caseData;
     }
 
-    private CaseData getCaseData(UUID caseUUID) {
+    protected CaseData getCaseData(UUID caseUUID) {
```
nickmcmahon01, line 131: We were auditing a case view here and we don't need to if we use the 'internal' method

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageResource.java
```diff
@@ -62,7 +62,15 @@ ResponseEntity<CreateStageResponse> createStage(@PathVariable UUID caseUUID, @Re
     @Authorised(accessLevel = AccessLevel.READ)
     @PutMapping(value = "/case/{caseUUID}/stage/{stageUUID}/recreate")
     ResponseEntity recreateStageTeam(@PathVariable UUID caseUUID, @PathVariable UUID stageUUID, @RequestBody RecreateStageRequest request) {
-        stageService.recreateStage(caseUUID, request);
+        stageService.createStage(caseUUID, new CreateStageRequest(
```
LiamMacP, line 65: Will this go through before the other change that removes this? This is now `@Deprecated` so I don't feel it needs changing as nothing will end up through this path.

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
```diff
@@ -136,101 +131,39 @@ StageWithCaseData getActiveStage(UUID caseUUID, UUID stageUUID) {
...
+        UUID currentUserUUID = allStages
+                .stream()
+                .map(BaseStage::getUserUUID)
+                .findFirst()
+                .orElse(null);
 
-        updateAssignmentAudit(stageToActivate.get());
+        UUID newTeamUUID = Objects.requireNonNullElse(
+                createStageRequest.getTeamUUID(),
+                infoClient.getTeamForStageType(createStageRequest.getType()).getUuid());
```
LiamMacP, line 145: What if there isn't one? Null would be returned and lead to a case without a active stage. 

We need to go back and look into this further. There are workflows such as POGR and MPAM that don't work on the principle of one team owns that stage. Which is by design for security. 

Do they need a overarching default team? Thats a question for another day.

src/main/java/uk/gov/digital/ho/hocs/casework/domain/repository/StageRepository.java
```diff
@@ -58,4 +59,7 @@ public interface StageRepository extends CrudRepository<BaseStage, Long> {
     @Query(value = "SELECT sd.* FROM stage_data sd where sd.case_reference = ?1", nativeQuery = true)
     Set<StageWithCaseData> findByCaseReference(String reference);
 
+    @Modifying
+    @Query(value = "UPDATE stage SET team_uuid = null, user_uuid= null WHERE case_uuid = ?1", nativeQuery = true)
+    void unAssignStagesForCase(UUID caseUUID);
```
LiamMacP, line 64:
```suggestion
    void unassignStagesForCase(UUID caseUUID);
```

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
```diff
@@ -612,7 +511,7 @@ private static Stream<StageWithCaseData> reduceToMostActive(List<StageWithCaseDa
 
     public void withdrawCase(UUID caseUUID, UUID stageUUID, WithdrawCaseRequest request) {
         log.info("About to withdraw case : {}", caseUUID);
-        CaseData caseData = caseDataService.getCase(caseUUID);
+        CaseData caseData = caseDataService.getCaseData(caseUUID);
```
LiamMacP, line 514: This would have been better under its own commit 🙂

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
```diff
@@ -136,101 +131,39 @@ StageWithCaseData getActiveStage(UUID caseUUID, UUID stageUUID) {
...
-        }
-
-        // Update audit and timeline unless active and allocated to same user and team - positioned here maintains timeline consistency.
-        if (isRecreation.get() && !isRecreateOfCurrentlyActive.get()) {
-            auditClient.recreateStage(stageToActivate.get());
-        } else if (!isRecreation.get()) {
-            auditClient.createStage(stageToActivate.get());
-        }
+        // Store the current user before clearing it later
+        UUID currentUserUUID = allStages
```
LiamMacP, line 137: This is only used in the assignment notification - if there is more than one stage active then this is flawed and could end up sending an email to the wrong person. 

Would a better solution be adding a check constraint on the stage table to ensure only 1 active stage?

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageResource.java
```diff
@@ -62,7 +62,15 @@ ResponseEntity<CreateStageResponse> createStage(@PathVariable UUID caseUUID, @Re
     @Authorised(accessLevel = AccessLevel.READ)
```
LiamMacP, line 62: Irrelavent to this PR but this should not be READ level.

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageResource.java
nickmcmahon01: it will if the BPMN commands it so? no?

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
nickmcmahon01: it was designed to support parallel stages, but the system has moved too far away in other areas, we're probably better off splitting the table into a stage and current_stage with the team,user and fk back to stage and case

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageResource.java
LiamMacP: This method isn't called anymore in workflow and the create endpoint is used in its lieu. This is only here for overhang at the minute and should be removed on the next `Major` change. 

This was only ever called by the `bpmnService` in workflow and was conditional.

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
nickmcmahon01: this is literally the 'bug' we've been having, there should be a team in stage in the diagrams and a 'fallback' team in the database for lost cases. there is no fallback stage for the TO cases at least.

LiamMacP: This is knowledge that was lost in the passage of time. And inflexible as only supports the DCU workflow. We probably need to look at a better solution.

nickmcmahon01: The original intent was that casework should be 'dumb'; so the calls to info are not in the right place and are the concern of the workflow service. moving them into casework hasn't solved anything wrt cases falling off the workstacks

LiamMacP: I disagree, Casework is the powerhouse as Workflow should be just a ACL of Camunda and shouldn't know anything about our implementation and should do basic things i.e. 'move forward', clear X'

PR #670, opened 2022-05-25
HOCS-3976: Remove 'case viewed' audit when not needed

nickmcmahon01:
HOCS-3976: Remove 'case viewed' audit in operations that are not visible to the user

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
@@ -128,7 +128,7 @@ public class CaseDataService {
         return caseData;
     }
 
-    private CaseData getCaseData(UUID caseUUID) {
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
```diff
@@ -128,7 +128,7 @@ public CaseData getCase(UUID caseUUID) {
         return caseData;
     }
 
-    private CaseData getCaseData(UUID caseUUID) {
+    public CaseData getCaseData(UUID caseUUID) {
```
LiamMacP, line 131: Wasn't this changed to `protected` in a inflight draft PR?

nickmcmahon01: 👍 ⚡

PR #684, opened 2022-06-17
HOCS-5240: allow return of deleted correspondents

LiamMacP:
Change to the GET correspondent endpoint to support an Optional request
param for getting all correspondents including deleted ones. As this is
optional the existing behaviour is default and anything requiring can
opt in.

This has caused a re-write and simplification of the Correspondents
objects and JPA implementation. It now defaults to using the
Correspondent and on the endpoint required it retrieves the case data
and mark the correct correspondent with the primary flag.


src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/Correspondent.java
```diff
@@ -1,20 +1,114 @@
...
 import javax.persistence.Table;
+import javax.persistence.Transient;
 import java.time.LocalDateTime;
 import java.util.UUID;
 
 @NoArgsConstructor(access = AccessLevel.PROTECTED)
 @Entity
 @Table(name = "correspondent")
-public class Correspondent extends BaseCorrespondent {
+@EqualsAndHashCode
```
nickmcmahon01, line 24: Always suspicious of this on entities, the identity is the UUID, not the value of all the fields.

nickmcmahon01 (approved):

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/Correspondent.java
LiamMacP: I was thinking that, tempted on just specifying just the UUID.

nickmcmahon01 (approved):

PR #686, opened 2022-06-24
HOCS-5251 : Remove actionDataUuid from Docs service

sameerinfo2003:
JIRA Link : Remove actionDataUuid from Docs service

Screenshot 2022-06-24 at 10 35 40


LiamMacP (approved):
Approved - looks fine - we should really have done a minor change before marking as `@Deprecated` but as we're the only users and its not totally working as expected this is fine.

PR #689, opened 2022-07-01
HOCS-5278: add field restriction list

LiamMacP:
This change causes fields to be removed from the case data on restricted
cases by exemption rather than an inclusion policy. This is done
through the `RestrictedField[Service/Repository]` instead of  behaviour
added previously to the UserPermissionService that communicated with
`hocs-info-service`.

As we are handling changes to collections whereby the objects within a
mutable - this change removes the need for helper static classes.


src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/RestrictedFieldPermissions.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.Map;
+import java.util.Set;
+
+public class RestrictedFieldPermissions {
+
+    @JsonValue
+    public Map<String, FieldAccessLevels> caseFieldAccessLevels;
+
+    @JsonCreator
```
nickmcmahon01, line 15: Thankyou for un-lombocking

src/main/resources/config/restricted-fields.json
```diff
@@ -0,0 +1,59 @@
+{
+  "SMC" : {
```
nickmcmahon01, line 2: I think this 'soft link' is fine looking through the document removes a load of UUIDs too

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/RestrictedFieldPermissions.java
LiamMacP: I had it with Lombok but JsonCreator actually needs an AllArgsConstructor to work

src/main/java/uk/gov/digital/ho/hocs/casework/api/dto/GetCaseSummaryResponse.java
```diff
@@ -93,28 +92,6 @@ public static GetCaseSummaryResponse from(CaseSummary caseSummary) {
                 caseSummary.getActions(),
                 caseSummary.getSuspended()
         );
-
-
-    }
-
-    protected void replaceAdditionalFields(List<AdditionalFieldDto> additionalFieldDtoList) {
```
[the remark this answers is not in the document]
LiamMacP: They were helpers to just clear out collections, a bit pointless when while the properties were immutable, the items in could just be wiped.

src/main/resources/config/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+```json
+{
+ <<CASE_TYPE>>: {
+  <<PERMISSION_LEVEL>>: [
+   <<FIELD_NAMES>>
+  ] 
+ }
+}
+```
```
nickmcmahon01, line 22: I'm new to the project, how to do I add one for a new case type?

nickmcmahon01 (approved):

src/main/resources/config/README.md
LiamMacP: https://cheatography.com/gaston/cheat-sheets/json/

PR #704, opened 2022-07-29
HOCS-4791: Add ability to create POGR2 cases

JosephRogerson:


src/main/java/uk/gov/digital/ho/hocs/casework/api/SomuItemService.java
```diff
@@ -104,4 +105,16 @@ public void deleteSomuItem(UUID somuItemUuid) {
...
+
+        Set<SomuItem> toSomuItems = getItemsByCaseUuid(fromCase);
+        Set<SomuItem> fromSomuItems = toSomuItems.stream().map(item -> addCaseSomuItemBySomuTypeUuid(toCase, item.getSomuUuid(), item.getData())).collect(Collectors.toSet());
+
+        if (toSomuItems.size() == fromSomuItems.size()) {
+            log.debug("Copied Somu Items from case: {} to case {}", fromCase, toCase);
+        } else {
+            log.debug("Failed to copy Somu Items from case: {} to case {}", fromCase, toCase);
+        }
+    }
```
robertdeniszczyc2, line 119: As I understand this we're copying the somu item data but not the underlying documents, so we'll have references to the documents but the documents themselves will then be shared across two cases and two case types? I'm not sure this is desirable as making a change on one case could impact the other - for example if we deleted the document on a case.

Might be worth another look at how we approach this, so we maintain exclusivity between cases. Although we're calling the behaviour copying cases/items, I think what we are more looking for is a creation of new data but with the same content as the source, if that makes sense?

Happy to talk through any questions, give me a shout

robertdeniszczyc2 (changes requested):

src/main/java/uk/gov/digital/ho/hocs/casework/api/SomuItemService.java
JosephRogerson: Following discussion on the ticket, we no longer need to copy somu items. I have removed this copy mechanism from the PR

robertdeniszczyc2 (approved):

PR #713, opened 2022-08-11
HOCS-4803: Automatically assign POGR2 cases

JosephRogerson:


src/main/java/uk/gov/digital/ho/hocs/casework/api/factory/strategies/CopyPOGRtoPOGR2.java
```diff
@@ -12,7 +12,7 @@
 public class CopyPOGRtoPOGR2 extends AbstractCaseCopyStrategy implements CaseCopyStrategy {
 
     private static final String[] DATA_CLOB_KEYS = {
-            "BusArea",
+            "BusinessArea",
```
LiamMacP, line 15: Do we need all of these copied across?

JosephRogerson: These were all the values that were specified in the AC for HOCS-4791. That AC has been dropped from that ticket  because it can't be tested at the moment, but it's being moved to another ticket so it's still required

LiamMacP (approved):

PR #728, opened 2022-09-01
HOCS-5545: refactor case summary

LiamMacP:
The summary view for a case currently calls out to `hocs-info-service` for the additional fields to show. This is causing unnecessary coupling between fields, screens, stages and case type. 

As part of the ongoing `hocs-data` deprecation, this change pulls out the fields that should be displayed in the summary for a case into a JSON configuration file. To support this a new folder reading scanner has been implemented that loops around files and builds representation that can be used.


src/main/java/uk/gov/digital/ho/hocs/casework/domain/repository/JsonConfigFolderReader.java
```diff
@@ -0,0 +1,67 @@
...
+public abstract class JsonConfigFolderReader {
+
+    private final ObjectMapper objectMapper;
+
+    protected JsonConfigFolderReader(ObjectMapper objectMapper) {
+        this.objectMapper = objectMapper;
+    }
+
+    protected <T extends CaseTypeObject<V>, V> Map<String, V> readValueFromFolder(TypeReference<T> reference) {
+        var folder = Thread.currentThread().getContextClassLoader().getResource(String.format("config/%s", getFolderName()));
```
LiamMacP, line 30: There is a `getResources` method to this that returns an Enumerable and likely could be used. While this code is likely less performant, we'd either have to do weird `while...do` logic or convert to a `EnumerationSpliterator`. 

I've chosen readability over simplicity/performance here.

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
```diff
@@ -550,49 +552,26 @@ void deleteCase(UUID caseUUID, Boolean deleted) {
...
-        CaseActionDataResponseDto caseActionData = caseActionService.getAllCaseActionDataForCase(caseUUID);
-
-        summaryBuilder.withActions(caseActionData);
-
-        Map<String, String> caseDataMap = caseData.getDataMap();
-        summaryBuilder.withAdditionalFields(getAdditionalFieldsForSummary(summaryFields, caseDataMap));
-        summaryBuilder.withStageDeadlines(getStageDeadlines(caseData, caseDataMap));
-
-
-        auditClient.viewCaseSummaryAudit(caseData);
```
LiamMacP, line 579: This is an important decision to make - is this useful when to view the case summary you have already effectively viewed a case. The only reason this might be good is that you can load a case but not necessarily look at the summary tab.

LiamMacP (commented):
These configuration files are pulled directly from the `hocs-info-service` call and then stripped down, we should get these reviewed in the future as there is potential with duplication of values. 

We may need to extend this to load choices for somu-lists/ multiple static lists per field.

src/main/java/uk/gov/digital/ho/hocs/casework/domain/repository/JsonConfigFolderReader.java
```diff
@@ -0,0 +1,71 @@
+package uk.gov.digital.ho.hocs.casework.domain.repository;
```
LiamMacP, line 1: It might be worth looking at whether we want to move this into a package that we ingest within the repos.

SWhitlam (approved):

PR #737, opened 2022-09-07
Add basic helm chart, replacing KD

nickmcmahon01:


.drone.yml
```diff
@@ -1,3 +1,29 @@
...
+steps:
+  - name: cs-dev
+    image: quay.io/ukhomeofficedigital/helm:v3.5.4
+    environment:
+      KUBE_CLUSTER_NAME: acp-notprod
+      KUBE_NAMESPACE: cs-dev
+      KUBE_SERVER: https://kube-api-notprod.notprod.acp.homeoffice.gov.uk
+      KUBE_TOKEN:
+        from_secret: cs_notprod
+      VALUES_FILE: '--values ./helm/values-notprod.yaml'
```
SWhitlam, line 20: I still think we should pass in a values-namespace.yaml file, maybe in additoin to the not-prod.yaml? I wonder if it's possible not fail if it doesn't exist?

nickmcmahon01: it will apply them in order so notprod first, then *cs-*.yaml

helm/hocs-casework/templates/_envs.tpl
```diff
@@ -0,0 +1,91 @@
+{{- define "deployment.envs" }}
+- name: JAVA_OPTS
```
zuzak, line 2: Use [YAML folding syntax](https://yaml-multiline.info/) to split this over many lines instead of one big long one.

PR #742, opened 2022-09-14
HOCS-5569: refactor read only accordion

LiamMacP:
Add the services and repository surrounding loading of the case
details (unallocated view) from an group/ stage configuration. As the
`group` way is what we are envisaging to move towards, this is an early
retrieval and exit. As current CS case types run on a stage based
system, the fallback is to look at the stage configuration folder and
return a dataset that is like the current version.


src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataDetailsService.java
```diff
@@ -0,0 +1,74 @@
...
+                    caseData.getReference(),
+                    caseDataGroupFields,
+                    caseData.getDataMap());
+        }
+
+        var caseDataStageFields = caseDataDetailsStagesRepository.getStagesDetailsFieldsByType(caseData.getType());
+        var caseStages = stageService.getAllStagesByCaseUUID(caseUuid).stream()
+                .map(Stage::getStageType).toList();
+
+        /*
```
LiamMacP, line 53: We should ensure a ticket exists for this TODO.

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/CaseDataDetails.java
```diff
@@ -0,0 +1,53 @@
...
+        return groups;
+    }
+
+    @Getter
+    public static class Fields {
+
+        private final String name;
+        private final String label;
+        private final String component;
+        private final Map<String, Object> props;
```
LiamMacP, line 39: The reason why props is here and not a slimmed down version is because we should support the ability for this to show the fields outside of the mapped-display component. Having props here (which we can cut down some of the values) - acts as a natural grouping.

src/main/java/uk/gov/digital/ho/hocs/casework/domain/repository/CaseDataDetailsGroupsRepository.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public final Map<String, Map<String, List<CaseDataDetails.Fields>>> detailsFields;
+
+    public CaseDataDetailsGroupsRepository(ObjectMapper objectMapper) {
+        super(objectMapper);
+
+        detailsFields = readValueFromFolder(new TypeReference<CaseDataDetails>() {});
+    }
+
+    public Map<String, List<CaseDataDetails.Fields>> getGroupsDetailsFieldsByType(String type) {
```
LiamMacP, line 29: We likely could be smarter with this if we wanted to be with generics and more this to the base class. Unsure if its worth the hassle though.

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataDetailsService.java
LiamMacP: This can be circled around on the same ticket - it's now blocked by UCD.

src/main/java/uk/gov/digital/ho/hocs/casework/domain/repository/CaseDataDetailsGroupsRepository.java
LiamMacP: This can be for a second pass if we want to - I'm not convinced of the benefit here.

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/CaseDataDetails.java
LiamMacP: I've left all of the props for now incase we want to do something fancy - we can always revisit and remove unnecessary ones.

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataDetailsService.java
```diff
@@ -0,0 +1,77 @@
...
+            return new CaseDataDetailsDto(caseData.getType(), caseData.getReference(), caseDataStageFields,
+                caseData.getDataMap());
+        }
+
+        var caseStages = stageService.getAllStagesByCaseUUID(caseUuid).stream().map(Stage::getStageType).toList();
+
+        /*
+         * TODO: Remove this call by bringing stage types into casework or reworking existing case type implementation
+         *  to run in the group format.
+         */
```
nickmcmahon01, line 59: 👍

LiamMacP:
> Once this gets merged, how do developers know how to use this? What is in flight (potentially) or in the sprint that will need updating?

I will plan a demo after merge on this - the fallout risk can be identified then. There is a weight of developers having to smoke test their changes before creating PRs against the AC.

nickmcmahon01:
👌

nickmcmahon01 (approved):

PR #745, opened 2022-09-22
HOCS-5585: Migration audit

rjweeks:
### HOCS-5585

Create the closed case migration audit events.

### DIFF
--- a/src/main/resources/application-local.properties
+++ b/src/main/resources/application-local.properties
@@ -22,6 +22,6 @@ aws.sns.audit-search.arn=arn:aws:sns:${aws.sns.config.region}:${aws.sns.audit-se
 
 spring.datasource.hikari.minimumIdle=2
 
-migration.userid=f55d4370-3955-44b8-91d0-f1b08c967a2b
+migration.userid=
 migration.username=migration
 migration.group=cms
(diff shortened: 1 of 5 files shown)


LiamMacP (changes requested):
I'd prefer for us to have a cut down `MigrationAuditClient` with just the methods we need for this to work as we can just add new ones as we need.

src/main/resources/application-local.properties
```diff
@@ -21,3 +21,7 @@ aws.sns.audit-search.topic-name=hocs-audit-topic
 aws.sns.audit-search.arn=arn:aws:sns:${aws.sns.config.region}:${aws.sns.audit-search.account.id}:${aws.sns.audit-search.topic-name}
 
 spring.datasource.hikari.minimumIdle=2
+
+migration.userid=f55d4370-3955-44b8-91d0-f1b08c967a2b
```
LiamMacP, line 25: Should this have been committed?

SWhitlam (approved):

src/main/resources/application-local.properties
rjweeks: It was randomly generated. I will remove it.

LiamMacP (approved):

PR #764, opened 2022-10-18
HOCS-5725: disable proxy image in dev only

nickmcmahon01:


helm/hocs-casework/Chart.yaml
```diff
@@ -3,5 +3,5 @@ name: hocs-casework
 version: 1.0.0
 dependencies:
   - name: hocs-generic-service
-    version: 1.0.0
+    version: 1.1.1
```
zuzak, line 6: Two jumps?
```suggestion
    version: 1.1.0
```

src/main/resources/application-tls.properties
```diff
@@ -0,0 +1,5 @@
+server.ssl.key-store-type=PKCS12
+server.ssl.key-store=file:${jdk.keystore.file}
+server.ssl.key-store-password=${jdk.keystore.password:changeit}
```
zuzak, line 3: If we can make the Helm chart autogenerate a password in a secret prior to deploy that'd be cool, but we should otherwise write it so that if jdk.keystore.password is undefined the app crashes rather than falling back to `changeit`?

zuzak (approved):

src/main/resources/application-tls.properties
nickmcmahon01: https://github.com/UKHomeOffice/cfssl-sidekick/blob/master/scripts/create-keystore.sh#L11 'changeit' is literally the password! it's not great is it. we could set up a better password in a secret and point this, and cfssl-sidekick at both, but I don't see an easy way for the sidekick to pick it up

PR #778, opened 2022-10-21
HOCS-5726: IEDET - Complaint sub categories are not displayed correctly in case data extracts

AlexW013:


robertdeniszczyc2:
There's a change to the helm-deploy on the file list but it doesn't show any diff, is the change intended?

PR #779, opened 2022-10-25
Migration case attachments

rjweeks:
### HOCS-5584

Add case attachments (documents) to Document Service.


src/main/java/uk/gov/digital/ho/hocs/casework/migration/api/dto/CaseAttachment.java
```diff
@@ -0,0 +1,23 @@
...
+@Getter
+public class CaseAttachment {
+
+    @JsonProperty("displayName")
+    private String displayName;
+
+    @JsonProperty("documentType")
+    private String type;
+
+    @JsonProperty("s3UntrustedUrl")
```
LiamMacP, line 21: As commented on the other one - I think this probably should be `documentUuid` or similar

SWhitlam: > As commented on the other one - I think this probably should be `documentUuid` or similar

I don't think this is the document UUID though? It's the URL of the file in the untrusted bucket which happens to be a UUID?

LiamMacP: `documentPath` then? Or even `fileLink` as you suggested? I don't see the value of having `s3Untrusted` nor `Url` as its not strictly accurate

rjweeks: Updated.

PR #781, opened 2022-10-26
HOCS-5690: add case data tag implementation

LiamMacP:
Add the case data tag table, this contains the case uuid and tag as the
primary composite key. Alongside this we store a created_at and
deleted_on timestamp to enable any potential future investigations. The
change also provides an index based on deleted_on being null to allow
for quick lookups.

Add E2E for adding and removing tags to a case.

On request for a stage the query the SQL now returns all the tags within
the newly created table. To ensure backwards compatibility while
refactoring works - the previous behaviour is maintained and is
appended into any tags created.


src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseTagResource.java
```diff
@@ -0,0 +1,39 @@
...
+@RestController
+class CaseTagResource {
+
+    private final CaseTagService caseTagService;
+
+    public CaseTagResource(CaseTagService caseTagService) {
+        this.caseTagService = caseTagService;
+    }
+
+    //@Authorised(accessLevel = AccessLevel.WRITE)
```
LiamMacP, line 25: I'd like for us to somehow be able to only allow this from internal services - through a new header maybe thats injected by workflow service?

This would bypass this authorisation check.

LiamMacP: I think theres a presumption that if a user can progress a case that they have permissions of WRITE or higher.

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseTagService.java
```diff
@@ -0,0 +1,26 @@
...
+import uk.gov.digital.ho.hocs.casework.domain.repository.CaseTagRepository;
+
+import java.util.UUID;
+
+@Service
+public class CaseTagService {
+
+    private final CaseTagRepository caseTagRepository;
+
+    public CaseTagService(final CaseTagRepository caseTagRepository) {
```
LiamMacP, line 14:
```suggestion
    public CaseTagService(CaseTagRepository caseTagRepository) {
```

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
```diff
@@ -529,7 +549,7 @@ private void updateStages(Set<StageWithCaseData> stages) {
         for (StageWithCaseData stage : stages) {
             stagePriorityCalculator.updatePriority(stage, stage.getCaseDataType());
             daysElapsedCalculator.updateDaysElapsed(stage.getData(), stage.getCaseDataType());
-            stage.setTag(stageTagsDecorator.decorateTags(stage.getData(), stage.getStageType()));
+            //stage.setTag(stageTagsDecorator.decorateTags(stage.getData(), stage.getStageType()));
```
LiamMacP, line 552: No longer required as retrieved from the table.

src/main/java/uk/gov/digital/ho/hocs/casework/api/dto/CaseTagDto.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Getter;
+import uk.gov.digital.ho.hocs.casework.domain.model.CaseDataTag;
+
+import java.time.LocalDateTime;
+import java.util.UUID;
+
+@Getter
+public class CaseTagDto {
```
LiamMacP, line 11: There is a question about whether we need this - but I feel that this is more of a RESTful way of handling this POST. We can then use the createdAt for the business event thats emitted.

src/main/java/uk/gov/digital/ho/hocs/casework/api/dto/GetStageResponse.java
```diff
@@ -90,12 +91,14 @@ public class GetStageResponse {
     private String nextCaseStageUUID;
 
     public static GetStageResponse from(StageWithCaseData stage) {
+        var tags =
```
LiamMacP, line 94: This has 0(n) complexity on the return for cases that have a tag - but none for all other cases.

This is an improvement other the case data lookup that happened in `decorateTags` before.

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/CaseDataTagId.java
```diff
@@ -0,0 +1,42 @@
+package uk.gov.digital.ho.hocs.casework.domain.model;
+
+import java.io.Serializable;
+import java.util.Objects;
+import java.util.UUID;
+
+public class CaseDataTagId implements Serializable {
```
LiamMacP, line 7: Due to the composite key - we likely could get around this with not having a composite key but feels a good place as enforces no duplicate tags.

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/StageWithCaseData.java
```diff
@@ -74,9 +82,10 @@ public class StageWithCaseData extends BaseStage {
 
     @Getter
     @Setter
-    @Transient
-    @JsonInclude(JsonInclude.Include.NON_NULL)
-    private ArrayList<String> tag;
+    @OneToMany(fetch = FetchType.EAGER)
+    @JoinColumn(name = "case_uuid", referencedColumnName = "case_uuid")
+    @Where(clause = "deleted_on IS NULL")
```
LiamMacP, line 87: This is a bit of a pain - if this was a native query it would be simpler. I think this is okay though...

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
LiamMacP: For the time being this has been reinstated but integrated into the functionality.

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseTagResource.java
LiamMacP: As this is for internal use this does not need an Authorised - we need to look at how we can protect internal APIs.

src/main/java/uk/gov/digital/ho/hocs/casework/domain/exception/ApplicationExceptions.java
```diff
@@ -138,4 +138,22 @@
 
     }
 
+    class DatabaseConflictException extends RuntimeException {
+        private final LogEvent event;
+        private final LogEvent exception;
+
+        public DatabaseConflictException(String msg, LogEvent event, Object... args) {
+            super(String.format(msg, args));
```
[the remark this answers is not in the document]
LiamMacP: This has been rectified.

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
```diff
@@ -529,7 +549,9 @@ private void updateStages(Set<StageWithCaseData> stages) {
         for (StageWithCaseData stage : stages) {
             stagePriorityCalculator.updatePriority(stage, stage.getCaseDataType());
             daysElapsedCalculator.updateDaysElapsed(stage.getData(), stage.getCaseDataType());
-            stage.setTag(stageTagsDecorator.decorateTags(stage.getData(), stage.getStageType()));
+
+            //TODO: Remove after Workflow Implementation released
```
robertdeniszczyc2, line 553: We should raise a separate ticket and reference the ticket in this if they are going separately

src/main/java/uk/gov/digital/ho/hocs/casework/application/LogEvent.java
```diff
@@ -1,153 +1,154 @@
 package uk.gov.digital.ho.hocs.casework.application;
```
robertdeniszczyc2, line 1: Is this change just reordering the list?

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/CaseDataTag.java
```diff
@@ -0,0 +1,80 @@
...
+import javax.persistence.Table;
+import java.io.Serializable;
+import java.time.LocalDateTime;
+import java.util.Objects;
+import java.util.UUID;
+
+@Entity
+@Table(name = "case_data_tag")
+@IdClass(CaseDataTagId.class)
+public class CaseDataTag implements Serializable {
```
robertdeniszczyc2, line 17: What would be the behaviour in a scenario where a tag is added, deleted, then added again? Do we need a differentiator between the two records stored in the table other than created_at and deleted_on?

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/CaseDataTagId.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private String tag;
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {return true;}
+        if (o == null || getClass() != o.getClass()) {return false;}
+        CaseDataTagId that = (CaseDataTagId) o;
+        return Objects.equals(caseUuid, that.caseUuid) && Objects.equals(tag, that.tag);
+    }
```
robertdeniszczyc2, line 19: This is essentially duplicated as in line 66 onwards in `src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/CaseDataTag.java` . Maybe some potential for reuse as it's a pretty generic method?

src/main/resources/db/migration/postgresql/V1_42__5690_ADD_CASE_DATA_TAG_TABLE.sql
```diff
@@ -0,0 +1,13 @@
+SET search_path TO casework;
+
+CREATE TABLE IF NOT EXISTS case_data_tag (
+    case_uuid UUID NOT NULL,
+    tag VARCHAR(20) NOT NULL,
```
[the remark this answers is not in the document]
robertdeniszczyc2: Makes sense to me - we should maybe somewhere defining allowed tags, but this feels like something maybe better handled at a workflow level as it's business logic essentially?

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/CaseDataTagId.java
LiamMacP: This is boilerplate equals implementation that IntelliJ gives you out the box. I'd prefer to keep the equals and hashcode stuff close to the class incase it ever needs to differ

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/CaseDataTag.java
LiamMacP: This doesn't work at all for tags that already exist and have been deleted, good catch...

src/main/java/uk/gov/digital/ho/hocs/casework/application/LogEvent.java
LiamMacP: `CASE_TAG_CONFLICT` has been added, apart from that yes

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageService.java
LiamMacP: This has been added

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/CaseDataTag.java
LiamMacP: I have altered this to work for deletes and re-adds

robertdeniszczyc2 (approved):

PR #797, opened 2022-11-17
HOCS-5503: Change IEDET categories screen

JosephRogerson:


LiamMacP (commented):
Is there a required change for hocs-audit aswell?

LiamMacP (approved):

JosephRogerson:
> Is there a required change for hocs-audit aswell?

No, we don't want to change what appears in extracts so the config there is already what we want

PR #802, opened 2022-11-23
HOCS-5946 Initial work to refactor workstacks to use JPA 

SWhitlam:
HOCS-5946 Initial work to refactor workstacks to use JPA read only queries to improve performance and move away from the StageWithCaseData view.

- Move Workstack specific code to new services and resources
- Use JPA Fetch queries instead of calling StageWithData View
- Use Workstack specific entities to enable using Fetch queries which require the @ID to be on the key used in foreign constraints
- Use CaseData as the root class rather than Stage - refactor of Contributions and Policies

Outstanding work:
- Case Link entity needs adding to CaseData
- Review test coverage to ensure we have adequate testing in place for workstacks
- performance smoke test to ensure it performs as expected in a prod like environment
- further refactoring, particularly of tests to make the code cleaner

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/casework/api/StageResource.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/casework/api/StageResource.java
@@ -2,14 +2,12 @@ package uk.gov.digital.ho.hocs.casework.api;
 
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.PutMapping;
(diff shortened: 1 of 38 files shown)


src/main/java/uk/gov/digital/ho/hocs/casework/api/StageResource.java
```diff
@@ -87,16 +84,6 @@ ResponseEntity updateStageUser(@PathVariable UUID caseUUID,
         return ResponseEntity.ok().build();
     }
 
-    @PutMapping(value = "/case/team/{teamUUID}/allocate/user/next")
```
LiamMacP, line 90: Is this unused or moved? There is a usecase from the frontend for it?

src/main/java/uk/gov/digital/ho/hocs/casework/api/StageResource.java
```diff
@@ -87,16 +84,6 @@ ResponseEntity updateStageUser(@PathVariable UUID caseUUID,
...
-    ResponseEntity allocateStageUser(@PathVariable UUID teamUUID,
```
LiamMacP, line 91: Is this unused too?

PR #809, opened 2022-12-01
Hocs 5566 rejected by psu tag

vinothon84:
HOCS-5566- Build changes to include validation if 
the tag already exist for a case and respective 
junit test been corrected.

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/casework/domain/repository/CaseTagRepository.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/casework/domain/repository/CaseTagRepository.java
@@ -18,8 +18,7 @@ public interface CaseTagRepository extends org.springframework.data.repository.R
     @Modifying
     CaseDataTag save(CaseDataTag entity);
 
-    @Query(value = "select  cdt.*  from  case_data_tag cdt where case_uuid = ?1  and tag = ?2", nativeQuery = true)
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/digital/ho/hocs/casework/domain/repository/CaseTagRepository.java
```diff
@@ -18,6 +18,9 @@ public interface CaseTagRepository extends org.springframework.data.repository.R
     @Modifying
     CaseDataTag save(CaseDataTag entity);
 
+    @Query(value = "select  cdt.*  from  case_data_tag cdt where case_uuid = ?1  and tag = ?2", nativeQuery = true)
```
LiamMacP, line 21: While we do use native queries over the product - I think the general direction is to move to a more JPA way of doing it. You should just need the method to be: 

`CaseDataTag findByCaseUuidAndTag(UUID caseUuid, String tag);`

Without the `Query`...

LiamMacP (changes requested):
Just one small alteration...

src/main/java/uk/gov/digital/ho/hocs/casework/domain/repository/CaseTagRepository.java
vinothon84: Incorporated and thanks for educating me on JPA side of things.

LiamMacP (approved):

PR #812, opened 2022-12-09
HOCS-5667: COMP add SMC radio option

JosephRogerson:


src/main/resources/config/summary/COMP.json
```diff
@@ -17,12 +17,17 @@
           "value": "Ex-Gratia"
         },
         {
-          "label": "Minor Misconduct",
```
LiamMacP, line 20: Should these be lower case - they feel like a proper noun to me? Resolve this commend if UCD have requested the lowercase 🙂

JosephRogerson: Sentence case has been requested here

LiamMacP (approved):

PR #823, opened 2022-12-19
HOCS-XXXX: handle null value in stage tag map

LiamMacP:
When querying the map for the stage tags, if there is a null value within the case data blob this triggers a NullPointerException. This change reworks the handling of this check and get to get the value and run a null check on it.


src/test/java/uk/gov/digital/ho/hocs/casework/api/StageTagsDecoratorImplTest.java
```diff
@@ -40,6 +40,12 @@ public void addsNoTagsWhenNotAHomeSecReplyStage() {
         assertEquals(0, tags.size());
     }
 
+    @Test
+    public void addsNoTagsWhenNullInMapAtCorrectStage() {
+        var tags = stageTagsDecorator.decorateTags(Map.of(HOME_SEC_REPLY_FIELD_NAME, null), "DCU_MIN_MARKUP");
+        assertEquals(0, tags.size());
+    }
+
```
robertdeniszczyc2, line 48: Not sure how "valid" (might be a better word) this test is as Map:
```
disallow null keys and values. Attempts to create them with null keys or values result in NullPointerException
```
https://docs.oracle.com/javase/9/docs/api/java/util/Map.html

We should instead look to defend against null being passed in to the Map, as I think even setting the behaviour to add no tags still leaves the potential for a NPE if `null` was passed into the map from the stage data?

Happy to talk through if I've misunderstood the intention.

LiamMacP: I think there are two parts to this: 

1. The underlying cause - we should be defensively ensuring that null is not added as its not valid in our context i.e. how did that null get in the case data blob.
2. The underlying result - we should be being defensive if a null gets into the map. We use HashMaps throughout the code and most of Spring Boot also utilises it. These by default do support null values: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/HashMap.html

LiamMacP: As discussed - updated the map used 🙂

src/test/java/uk/gov/digital/ho/hocs/casework/api/StageTagsDecoratorImplTest.java
```diff
@@ -40,6 +41,14 @@ public void addsNoTagsWhenNotAHomeSecReplyStage() {
         assertEquals(0, tags.size());
     }
 
+
+    @Test
+    public void addsNoTagsWhenNullInMapAtCorrectStage() {
+        var tags = stageTagsDecorator.decorateTags(Collections.singletonMap(HOME_SEC_REPLY_FIELD_NAME, null), "DCU_MIN_MARKUP");
+        assertEquals(0, tags.size());
+
```
robertdeniszczyc2, line 49:
```suggestion
```

robertdeniszczyc2 (approved):

PR #825, opened 2022-12-19
HOCS-5610: change search url in helm

LiamMacP:
Instead of using `hocs-search` for provided data on search, this change points the service to the `hocs-search-opensearch` service.


nickmcmahon01 (approved):
We can do this, or we can edit the hocs-search service in the cluster temporarily during the cut-over, it's up to you.

LiamMacP:
> We can do this, or we can edit the hocs-search service in the cluster temporarily during the cut-over, it's up to you.

I edited it in the cluster, as the OpenSearch move is done for DEV now, this brings it in line with where it should be.

PR #853, opened 2023-01-26
HOCS-6183: Remove incorrect value from COMP accordion

JosephRogerson:


jeff-horton-ho-sas (approved):
👍 - Needs semantic version label.

PR #871, opened 2023-02-10
HOCS-6196 Alter contribution processor logging

robertdeniszczyc2:
Move processor logging to DEBUG, introduce new INFO log line. New INFO level log will only be logged once for the set, rather than once per iteration as with the now DEBUG level log lines.


LiamMacP (approved):
Code change looks good - I think the CI update is okay but might be worth looking at

robertdeniszczyc2:
> Code change looks good - I think the CI update is okay but might be worth looking at

yeah I forgot to add the label, added and all green now so will merge

PR #883, opened 2023-02-20
HOCS-6225 - Add date_completed to case_data

jeff-horton-ho-sas:


src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/AbstractCaseData.java
```diff
@@ -101,6 +101,11 @@ public class AbstractCaseData implements Serializable {
     @Column(name = "completed")
     private boolean completed;
 
+    @Setter
+    @Getter
+    @Column(name = "date_completed")
+    private LocalDateTime dateCompleted;
```
LiamMacP, line 107: I'd prefer for this to replace the 'completed' flag eventually. Not blocking this but could we have have a ticket created to discuss this and any further actions? 🙂

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
```diff
@@ -574,9 +575,14 @@ void updatePrimaryTopic(UUID caseUUID, UUID stageUUID, UUID primaryTopicUUID) {
     }
 
     void completeCase(UUID caseUUID, boolean completed) {
+        completeCase(caseUUID, completed, completed ? LocalDateTime.now() : null);
```
nickmcmahon01, line 578: Does this localDatetime.now() also make its way through to audit service? I'm not sure what the right behaviour is to be honest.

src/main/resources/db/migration/postgresql/R__6225_create_open_cases_view.sql
```diff
@@ -0,0 +1,16 @@
+SET search_path TO casework;
+
+DROP VIEW IF EXISTS report_open_cases;
+
+CREATE OR REPLACE VIEW report_open_cases AS
+    SELECT cd.uuid                        AS caseUUID,
+           cd.data::json ->> 'BusArea'    AS business_area,
```
nickmcmahon01, line 7: Do you need this cast? you can just do cd.data->>'BusArea'

src/main/java/uk/gov/digital/ho/hocs/casework/api/CaseDataService.java
jeff-horton-ho-sas: Looking at this in a bit more detail I'm not sure the existing audit functionality is correct. 

Complete case is called by the stage service when a case is withdrawn with completed always being true, and directly from a REST endpoint where the completed flag comes from the request body. Note migration handling an already completed case is separate (https://github.com/UKHomeOffice/hocs-casework/pull/883/files#diff-89b3d3c138306f66d3069d413b400189545aff2a193f134d16f47e642c1f02d4R115)  - but also has issues that I've raised with Blaine. @SWhitlam also mentioned there were some changes planned in that area so it may already be in hand.

If completed is false then the current stage is still completed (but setting `caseData.update(CaseworkConstants.CURRENT_STAGE, "");` is conditional on completed being true) And the COMPLETE_CASE audit event is still sent with just the case_uuid as the payload so no way to determine if it was in fact un-completing the case. 

Possibly worth a discussion on what the behaviour should be. Thoughts:
* It seems like if un-completing a case is possible, it needs additional updates e.g. assigning the case to a stage. The endpoint and method should only be for completing a case and undoing that should be split into its own implementation if needed?
* `LocalDateTime.now()` is fine here as it's intended for completing the case right now. Alternatively we could accept the time from the rest endpoint but I don't see a use case for using anything other than now, esp. as migration uses a different endpoint anyway.
* We need a wider conversation about how we're logging audit events for migrated cases if it hasn't already happened. It seems wrong to log those events with the migration timestamp. May already be being addressed though.

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/AbstractCaseData.java
jeff-horton-ho-sas: created [HOCS-6410](https://collaboration.homeoffice.gov.uk/jira/browse/HOCS-6410)

PR #884, opened 2023-02-20
HOCS-6234: Add migration document to tags.

rjweeks:


src/main/resources/config/document-tags/BF.json
```diff
@@ -9,5 +9,6 @@
     "CRF",
     "DRAFT",
     "Final response"
```
reub-the-cube, line 11: Comma!

rjweeks: Fixed.

SWhitlam (approved):

PR #900, opened 2023-02-28
[HOCS-6266][Spike] Augment report data using hocs-info-service

jeff-horton-ho-sas:
https://collaboration.homeoffice.gov.uk/jira/browse/HOCS-6266


jeff-horton-ho-sas:
Includes changes from #883

src/main/java/uk/gov/digital/ho/hocs/casework/reports/api/dto/ReportColumnDto.java
```diff
@@ -0,0 +1,49 @@
...
+    @JsonProperty("render_in_csv")
+    @NonNull
+    private Boolean renderInCSV;
+    @JsonProperty("additional_fields")
+    private Map<String, String> additionalFields;
+    @JsonProperty("filter_type")
+    private FilterType filterType;
+
+    public ReportColumnDto withAdditionalField(String key, String value) {
+        if(additionalFields == null) {
```
SWhitlam, line 37: This could be initialised in the constructor (rathe than using Lombok)

src/main/java/uk/gov/digital/ho/hocs/casework/reports/domain/mapping/OpenCasesDataMapper.java
```diff
@@ -0,0 +1,52 @@
...
+
+    private final UserNameValueMapper userNameValueMapper;
+
+    private final TeamNameValueMapper teamNameValueMapper;
+
+    private final StageNameValueMapper stageNameValueMapper;
+
+    private final ExemptionDatesAgeAdjustmentLookup exemptionDatesAgeAdjustmentLookup;
+
+    @Autowired
```
SWhitlam, line 23: Autowired shouldn't be required here because the class is annotated with a Stereotype @Service

src/main/java/uk/gov/digital/ho/hocs/casework/reports/domain/mapping/UserNameValueMapper.java
```diff
@@ -0,0 +1,65 @@
...
+        refreshCache();
+    }
+
+    private String fetchUserNameByUUID(UUID userUUID) {
+        try {
+            UserDto user = infoServiceClient.getUser(userUUID);
+            return formatUsername(user);
+        } catch (Exception e) {
+            log.warn("Failed to fetch user with UUID {}", userUUID, REPORT_MAPPER_USER_CACHE_ERROR);
+            return null;
```
SWhitlam, line 42: I wonder if we would be better returning the UUID here? Returning null might imply the case isn't assigned to a user so while showing the user the UUID isn't ideal it at least maintains the expected functionality.

SWhitlam (commented):
It would be good to add some integration tests for the report endpoint to provide some end to end testing of the new API endpoints.

SWhitlam (approved):

PR #945, opened 2023-05-11
HOCS-6588 : COMP - Directorate selection doesn't match the value disp…

sameerinfo2003:
JIRA Link : https://collaboration.homeoffice.gov.uk/jira/browse/HOCS-6588

### DIFF
--- a/src/main/resources/config/details/stages/COMP.json
+++ b/src/main/resources/config/details/stages/COMP.json
@@ -593,7 +593,7 @@
                 },
                 {
                   "label": "EUSS FP",
-                  "value": "EUSSFP"
+                  "value": "EUSS FP"
                 }
               ],
               "conditionPropertyName": "OwningCSU",
@@ -1467,7 +1467,7 @@
                 },
                 {
                   "label": "EUSS FP",
-                  "value": "EUSSFP"
+                  "value": "EUSS FP"
                 }
               ],
               "conditionPropertyName": "OwningCSU",
@@ -2513,7 +2513,7 @@
                 },
                 {
                   "label": "EUSS FP",
-                  "value": "EUSSFP"
+                  "value": "EUSS FP"
                 }
               ],
               "conditionPropertyName": "OwningCSU",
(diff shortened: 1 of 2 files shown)


robertdeniszczyc2:
Hi @sameerinfo2003 ,

I've tested this locally and it doesn't seem to have the desired effect unless I have missed something. When I have tested it, the original EUSSFP text can still be seen in the summary and accordion, it isn't being replaced, so there is potentially extra changes required here to alter the display value.

Could you have a look to try to replicate please? Happy to jump on a call if needed to show what I see,

Thanks,
Rob

jeff-horton-ho-sas:
@sameerinfo2003 It looks like hocs-casework is still configured to show directorate options based on OwningCSU, which was removed here: https://collaboration.homeoffice.gov.uk/jira/browse/HOCS-6542. The hocs-casework props should match the hocs-workflow props, so the same change will need to be made in hocs-casework.

LiamMacP:
> @sameerinfo2003 It looks like hocs-casework is still configured to show directorate options based on OwningCSU, which was removed here: https://collaboration.homeoffice.gov.uk/jira/browse/HOCS-6542. The hocs-casework props should match the hocs-workflow props, so the same change will need to be made in hocs-casework.

They shouldn't always necessarily match, I don't know the context of the ticket, but as some cases have old values we need to support that within the accordion and summary.

jeff-horton-ho-sas:
> > @sameerinfo2003 It looks like hocs-casework is still configured to show directorate options based on OwningCSU, which was removed here: https://collaboration.homeoffice.gov.uk/jira/browse/HOCS-6542. The hocs-casework props should match the hocs-workflow props, so the same change will need to be made in hocs-casework.
> 
> They shouldn't always necessarily match, I don't know the context of the ticket, but as some cases have old values we need to support that within the accordion and summary.

Good point. In this case the mappings are still conditional on the OwningCSU value that has been removed. The hocs-casework props need to be changed to a single mapping of values to labels matching the current hocs-workflow props.

PR #951, opened 2023-05-18
[HOCS-6318] Return 409 Conflict on duplicate migrated reference

jeff-horton-ho-sas:


src/main/java/uk/gov/digital/ho/hocs/casework/domain/repository/CaseDataRepository.java
```diff
@@ -25,4 +26,7 @@ public interface CaseDataRepository extends CrudRepository<CaseData, Long> {
     @Query(value = "SELECT cd.* FROM case_data cd WHERE cd.uuid = ?1", nativeQuery = true)
     CaseData findAnyByUuid(UUID uuid);
 
+    @Query(value="SELECT cast(cd.uuid as text) FROM case_data cd WHERE cd.migrated_reference = ?1 LIMIT 1", nativeQuery = true)
```
LiamMacP, line 29: Happy to be wrong but do we need to do this cast? I'd have thought/ hoped it could do this internally

LiamMacP, line 29: Is there a check on the database column to ensure that only one case can have the same migrated_reference? If not, should there be? It will remove the need for the LIMIT 1.

src/main/java/uk/gov/digital/ho/hocs/casework/domain/repository/CaseDataRepository.java
```diff
@@ -25,4 +26,7 @@ public interface CaseDataRepository extends CrudRepository<CaseData, Long> {
...
+    Optional<UUID> findUUIDByMigratedReference(String migratedReference);
```
LiamMacP, line 30: Clever use of Optional - happy to let it through although not sure it adds much here

LiamMacP (approved):
Just a couple of questions, nothing blocking really

src/main/java/uk/gov/digital/ho/hocs/casework/domain/repository/CaseDataRepository.java
jeff-horton-ho-sas: It can't be enforced in the database as E.g. UAT needs to support multiple migrated references as the allow duplicates flag is on there. There are existing QA cases with duplicates so this is needed to ensure a 409 is thrown rather than a 500 when Hibernate throws an exception due to unexpected multiple results.

jeff-horton-ho-sas: The postgres dialect can't convert from the raw UUID representation in the db, throws "No Dialect mapping for JDBC type: 1111" exception. (where 1111 is a generic "OTHER" type that applies to UUIDs, JSONB, etc.)

There might be some tech debt work here to setup the dialect so it does just work, but that seems outside the scope of this feature.

PR #966, opened 2023-06-14
[HOCS-6385] return the created topic from addTopicToCase

jeff-horton-ho-sas:
https://collaboration.homeoffice.gov.uk/jira/browse/HOCS-6385

Case creator needs to add the topic, then set that topic to be the primary, which requires the UUID of the created topic.


jeff-horton-ho-sas:
Closing - requires active stage to create topic this way so needs a custom migration endpoint

PR #992, opened 2023-09-11
Hocs 6410 remove date completed column

vinothon84:
HOCS-6410: This PR will include changes to remove completed column from
casework views and remove them from both where clause wherever applicable.
Corresponding code changes were removed from hocs-casework - DTO, pojos
and junit test suits. This has been tested by making sure search functionality 
works.


src/main/resources/db/migration/postgresql/R__6225_active_case_view.sql
```diff
@@ -4,7 +4,7 @@ DROP VIEW IF EXISTS active_case;
 
 CREATE VIEW active_case
```
jeff-horton-ho-sas, line 5:
```suggestion
CREATE OR REPLACE VIEW active_case
```
We should be consistent with how we update the views. Remove the `DROP VIEW IF EXISTS active_case;` as well.

src/main/resources/db/migration/postgresql/R__6225_open_cases_report_view.sql
```diff
@@ -1,6 +1,6 @@
 SET search_path TO casework;
 
-DROP VIEW IF EXISTS report_open_cases;
+--DROP VIEW IF EXISTS report_open_cases;
```
jeff-horton-ho-sas, line 3:
```suggestion

```
No need to leave this commented out, we have it in version control

src/main/resources/db/migration/postgresql/R__6410_active_stage.sql
```diff
@@ -0,0 +1,22 @@
+SET search_path TO casework;
+
+--DROP VIEW IF EXISTS active_stage CASCADE;
```
jeff-horton-ho-sas, line 3:
```suggestion

```

src/main/resources/db/migration/postgresql/R__6410_active_stage.sql
```diff
@@ -0,0 +1,22 @@
...
+       s.transition_note_uuid,
+       s.case_uuid,
+       s.team_uuid,
+       s.user_uuid,
+       s.deadline_warning
+FROM casework.stage s
+WHERE s.team_uuid IS NOT NULL;
+
+alter table active_stage
+    owner to root;
```
jeff-horton-ho-sas, line 22: I'm not sure why this needs to be changed, it doesn't use `completed`. We shouldn't need this migration at all.

src/main/resources/db/migration/postgresql/R__6410_correspondents_json_by_case.sql
```diff
@@ -0,0 +1,27 @@
+SET search_path TO casework;
```
jeff-horton-ho-sas, line 1: I'm not sure why this needs to be changed, it doesn't use completed. We shouldn't need this migration at all.

src/main/resources/db/migration/postgresql/R__6410_open_cases_report_view.sql
```diff
@@ -0,0 +1,23 @@
...
+           ast.uuid                       AS stage_uuid,
+           ast.type                       AS stage_type,
+           ast.user_uuid                  AS assigned_user_uuid,
+           ast.team_uuid                  AS assigned_team_uuid,
+           cd.type                        AS case_type,
+           case_deadline < now()::date    AS outside_service_standard
+    FROM case_data cd
+    LEFT JOIN active_stage ast ON cd.uuid = ast.case_uuid
+    WHERE date_completed is null;
+
```
jeff-horton-ho-sas, line 23: This should not need a new migration, instead edit the existing file `R__6225_open_cases_report_view.sql` to use `date_completed`

src/test/java/uk/gov/digital/ho/hocs/casework/domain/model/CaseDataTest.java
```diff
@@ -32,7 +32,6 @@ public void getCaseData() {
         assertThat(caseData.getPrimaryTopicUUID()).isNull();
         assertThat(caseData.getCaseDeadline()).isNull();
         assertThat(caseData.getDateReceived()).isEqualTo(caseReceived);
-        assertThat(caseData.isCompleted()).isEqualTo(false);
```
jeff-horton-ho-sas, line 35: This and the other modified tests in this file should be replaced with a check that date completed has been set.

src/test/java/uk/gov/digital/ho/hocs/casework/migration/api/MigrationCaseDataServiceTest.java
```diff
@@ -125,7 +125,6 @@ public void shouldCreateOpenMigrationCase() throws ApplicationExceptions.EntityC
             originalReceivedDate, null, null, originalCreatedDate, migratedReference);
 
         // then
-        assertThat(caseData.isCompleted()).isEqualTo(false);
```
jeff-horton-ho-sas, line 128: This and the other modified tests in this file should be replaced with a check that date completed has been set.

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/AbstractCaseData.java
```diff
@@ -100,10 +100,6 @@ public class AbstractCaseData implements Serializable {
     @Column(name = "date_received")
     private LocalDate dateReceived;
 
-    @Setter
-    @Getter
-    @Column(name = "completed")
-    private boolean completed;
```
jeff-horton-ho-sas, line 106: This is being removed from the top level DTO but there is no migration to remove it from the `case_data` table.

src/main/java/uk/gov/digital/ho/hocs/casework/api/dto/GetCaseResponse.java
```diff
@@ -62,9 +62,6 @@ public class GetCaseResponse {
     @JsonProperty("stages")
     private List<SimpleStageDto> stages;
 
-    @JsonProperty("completed")
```
jeff-horton-ho-sas, line 65: We should maintain the same external API so there needs to be a getter to replace this.
```
@JsonProperty("completed")
public Boolean getCompleted() {
    return dateCompleted != null;
}
```
This should be included in tests as well.

jeff-horton-ho-sas (changes requested):
Great work. A few things to address, but otherwise good to go.

src/main/java/uk/gov/digital/ho/hocs/casework/domain/model/AbstractCaseData.java
vinothon84: I will include the completed column and will remove this as part of future migration (when we drop the column from case_data).

src/main/java/uk/gov/digital/ho/hocs/casework/api/dto/GetCaseResponse.java
```diff
@@ -62,8 +63,8 @@ public class GetCaseResponse {
...
-    private Boolean completed;
+    @JsonProperty("dateCompleted")
+    private LocalDateTime dateCompleted;
```
jeff-horton-ho-sas, line 67: It is good that we're adding dateCompleted to the response, but we still need to keep completed in the response JSON as well until clients of the API have been updated to use dateCompleted. This can be done by adding a getCompleted method with `@JsonProperty("completed")`:

```
@JsonProperty("completed")
public Boolean getCompleted() {
    return dateCompleted != null;
}
```

jeff-horton-ho-sas (changes requested):
One comment to resolve, otherwise LGTM 🙂.

src/main/java/uk/gov/digital/ho/hocs/casework/api/dto/GetCaseResponse.java
```diff
@@ -62,8 +64,8 @@ public class GetCaseResponse {
     @JsonProperty("stages")
     private List<SimpleStageDto> stages;
 
-    @JsonProperty("completed")
-    private Boolean completed;
+    @JsonProperty(value="completed", access=READ_ONLY)
+    private LocalDateTime dateCompleted;
```
jeff-horton-ho-sas, line 68: This is still not providing backwards compatibility for consumers of the existing api. Previously making a request to  `/case/data/{reference}` would have responded with an object including: 
```json
{
  // ... more fields
  "completed": true,
  // ... more fields
}
```
but now it will return something like:

```json
{
  // ... more fields
  "completed": "2023-09-12T01:02:03.456Z",
  // ... more fields
}
```

This will cause errors in consumers of the api, so we still need to provide completed as a boolean to existing API clients that expect it to be there. This ticket is concerned about there being a single source of truth in the DB for if a case is completed or not, not about also updating each service to switch to the new field right away.


```suggestion
    @JsonProperty("dateCompleted")
    private LocalDateTime dateCompleted;
    
    @JsonProperty(value="completed", access=READ_ONLY)
    public Boolean getCompleted() {
        return dateCompleted != null;
    }
```

jeff-horton-ho-sas (changes requested):

src/main/java/uk/gov/digital/ho/hocs/casework/api/dto/GetCaseResponse.java
vinothon84: I have incorporated above review comments and also added test cases to depend on dateCompleted.

jeff-horton-ho-sas (approved):

PR #998, opened 2023-09-19
HOCS-6741: build changes to remove completed column from db and pojos

vinothon84:
HOCS-6741: build changes to remove completed column from db and pojo.


jeff-horton-ho-sas (approved):
LGTM 🙂

PR #1031, opened 2025-07-03
[HOCS-7026] Add confirmation of sending reply to DCU for TRO cases

jeff-horton-ho-sas:
https://collaboration.homeoffice.gov.uk/jira/browse/HOCS-7026


mohomeoffice (approved):
looks okay to me

